Issue #1, opened 2018-04-01
Evaluate for release


jsumners:
@pinojs/pino please evaluate this codebase for release readiness. I'd like at least @mcollina and @davidmarkclements to sign off before I publish it.

@hbarcelos please verify that error stack traces are formatted correctly.

@ilyaztsv please verify that it properly supports defining which properties of an error object to display.

@victor0801x please verify that it outputs localized times as desired.

ilyaztsv:
Hi @jsumners !

I've checked `errorProps` flag and it works as expected.

But one thing, if I use the the flag without arguments (`pino-pretty --errorProps`) there is no error.
`Pino` provides the following error `errorProps flag provided without a string argument` in this case.

ilyaztsv:
Also one thing, related to https://github.com/pinojs/pino/issues/382
Here we still have the same issue. May be need to fix it after merging https://github.com/pinojs/pino/pull/383

jsumners:
@davidmarkclements:
+ Weird. I do not have any failing tests over here and I have the same environment. But this comment did make me realize the `--color` option would not pass in correctly. I have updated that to `--colorize`.
+ The idea is that `pino` will not ship with a prettifier. Instead, it will advertise those that exist. The constant need for more and more options to the prettifier, and the resulting dependency needs, are exactly the reason for it. Currently, the heaviest dependencies come from having to deal with locale data regarding date/time formatting. The way this is done in `pino`, at the moment, is way too fragile and non-standard (how do the format strings even work in that implementation?).

@ilyaztsv the exact same solution proposed in `pino` issue 383 is used here : https://github.com/pinojs/pino-pretty/blob/4e337de675f2984bb15cbac9d5742bd30518d61a/index.js#L215

Where are you seeing the `null` stack problem?

Hmm, I suppose it wouldn't be too difficult to add in an error if the the `--errorProps` flag is supplied and set to default.

jsumners:
@davidmarkclements try the tests again. I pushed some fixes.

davidmarkclements:
@jsumners 

* test now passing
* ok so it doesn't come bundled. 🔝💯

davidmarkclements:
LGTM - @mcollina over to you.

hbarcelos:
## Error stacktraces

LGTM.

---

## Formatting

Output on current prettifier on monorepo:
```
[2018-04-02T17:52:21.736Z] INFO (12446 on henrique-7650): Application is listening on http://localhost:8888
[...]

PR #3, opened 2018-04-04
fix: "will format date to local time" test

ilyaztsv:
Hello!

Related to https://github.com/pinojs/pino-pretty/issues/1#issuecomment-378203261 and 
screen shot 2018-04-04 at 20 28 39

I've fixed `will format date to local time` test.

### DIFF
--- a/test/basic.test.js
+++ b/test/basic.test.js
@@ -126,18 +126,21 @@ test('basic prettifier tests', (t) => {
       write (chunk, enc, cb) {
         const formatted = pretty(chunk.toString())
         const zonedDateTime = joda.ZonedDateTime.ofInstant(
-          joda.Instant.now(),
+          joda.Instant.ofEpochMilli(Date.now()),
           joda.ZoneOffset.SYSTEM
         )
         const offset = joda.DateTimeFormatter.ofPattern('Z').format(
           zonedDateTime
         )
-        const hour = Math.floor(
-          ((17 * 3600) + joda.ZoneOffset.of(offset).totalSeconds()) / 3600
+        const dateStr = joda.DateTimeFormatter.ofPattern('yyy-MM-dd').format(
+          zonedDateTime
+        )
+        const hour = joda.DateTimeFormatter.ofPattern('HH').format(
+          zonedDateTime
         )
         t.is(
           formatted,
-          `[2018-03-30 ${hour}:35:28.992 ${offset}] INFO (${pid} on ${hostname}): foo\n`
+          `[${dateStr} ${hour}:35:28.992 ${offset}] INFO (${pid} on ${hostname}): foo\n`
         )
         cb()
       }


jsumners:
Thank you. I'll merge once the CI tests finish.

jsumners:
It should look a little something like this: https://runkit.com/embed/0e4w2spy7r4u

ilyaztsv:
@jsumners thanks!

added:

```
const dateStr = joda.DateTimeFormatter.ofPattern('yyy-MM-dd').format(
          zonedDateTime
        )
```

test/basic.test.js
```diff
@@ -126,18 +126,21 @@ test('basic prettifier tests', (t) => {
...
+          joda.Instant.ofEpochMilli(Date.now()),
           joda.ZoneOffset.SYSTEM
         )
         const offset = joda.DateTimeFormatter.ofPattern('Z').format(
           zonedDateTime
         )
-        const hour = Math.floor(
-          ((17 * 3600) + joda.ZoneOffset.of(offset).totalSeconds()) / 3600
+        const dateStr = joda.DateTimeFormatter.ofPattern('yyy-MM-dd').format(
+          zonedDateTime
```
jsumners: This, and the hour, should be calculated from the static `epoch` variable. Otherwise, this test is always going to fail because `Date.now()` will always be after the static test data.

jsumners: This should also be `yyyy-MM-dd` (you're missing a `y`).

FYI: I originally calculated from the current `Date.now()` because I thought all I needed was the offset. That is not the case.

ilyaztsv: Not enough clear for me. `zonedDateTime` is create using 
```
joda.ZonedDateTime.ofInstant(
          joda.Instant.ofEpochMilli(Date.now()),
          joda.ZoneOffset.SYSTEM
        )
```

where `Date.now` is mocked using 'epoch` before each test:

```
t.beforeEach((done) => {
    Date.originalNow = Date.now
    Date.now = () => epoch

    done()
  })
```

So, We create zoned instant from `epoch = 1522431328992` and only format it (`dateStr` and `hour`).

Do you think that need change `Date.now` -> `epoch` here `joda.Instant.ofEpochMilli(Date.now())` ?

jsumners: I did forget that `Date.now` is mocked. However, it is mocked so that `pino` returns the correct time. Within the test itself I think it will be clearer to use `epoch`. When I read `Date.now()` I think "okay, the date at the exact time this statement is evaluated" where as if I read a locally defined constant then I know it's just that, a constant.

jsumners: In short, yes, `joda.Instant.ofEpochMilli(Date.now())` should be `joda.Instant.ofEpochMilli(epoch)` in the test.

jsumners:
Thank you for spending time on this. Given the discussion in #1 I have removed the `js-joda` library and replaced it with `dateformat`. While doing so, I believe I have fixed this bug.

PR #4, opened 2018-04-04
Alternative implementation of the meta wrapper

mcollina:
@jsumners @davidmarkclements let me know

### DIFF
--- a/index.js
+++ b/index.js
@@ -76,20 +76,9 @@ module.exports = function prettyFactory (options) {
     color.message = ctx.cyan
   }
 
-  if (!opts.useMetadata) {
-    return pretty
-  }
+  pretty.asMetaWrapper = asMetaWrapper
 
-  const outputSym = Symbol('pino-pretty.out')
-
-  return {
-    [Symbol.for('needsMetadata')]: true,
-    [outputSym]: opts.outputStream,
-    write (chunk) {
-      const formatted = pretty(chunk)
-      this[outputSym].write(formatted)
-    }
-  }
+  return pretty
 
   function pretty (inputData) {
     let log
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -249,3 +238,39 @@ module.exports = function prettyFactory (options) {
...
+    write (chunk) {
+      var chindings = this.lastLogger[parsed]
+
+      if (!chindings) {
+        chindings = JSON.parse('{"v":1' + this.lastLogger.chindings + '}')
+        this.lastLogger[parsed] = chindings
+      }
+
+      // TODO remove Object.assign as its slow
+      const obj = Object.assign({
```
jsumners: I tend to use https://www.npmjs.com/package/conflate when I'm looking for something other than `Object.assign`. See https://github.com/jsumners/merge-benchmarks

test/basic.test.js
```diff
@@ -354,7 +354,7 @@ test('basic prettifier tests', (t) => {
         cb()
       }
     })
-    const prettyStream = prettyFactory({useMetadata: true, outputStream: dest})
+    const prettyStream = prettyFactory().asMetaWrapper(dest)
```
jsumners: I'm confused here. How will options like `colorize` be recognized?

jsumners: Oh, I see how it works now. Can you add a simple unit test to show options being passed in? I think `levelFirst` would be sufficient.

index.js
mcollina: probably we should just leave `Object.assign`, the performance is not that horrible compared to others.

jsumners: I am 👌 with `Object.assign`.

mcollina:
Updated.

jsumners:
Just missing the readme update.

PR #8, opened 2018-04-08
translateTime accept an optional format string

victor0801x:
remove `--dateFormat` and `--localTime` options
`--translateTime` (`-t`) accept an optional format string
  - `-t` convert into UTC ISO format, default: `yyyy-mm-dd HH:MM:ss.l o`
  - `-t 'HH:MM:ss.l'` convert into UTC in 'HH:MM:ss.l' format
  - `-t 'SYS:standard'` convert into system's timezone, default: `yyyy-mm-dd HH:MM:ss.l o`
  - `-t 'SYS:HH:MM:ss'` convert into system's timezone in 'HH:MM:ss' format

#### checklist
- [x] `npm test` passes
- [x] tests and/or benchmarks are included
- [x] documentation is changed or added

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -48,10 +48,10 @@ error like objects. Default: `err,error`.
 + `--messageKey` (`-m`): Define the key that contains the main log message.
 Default: `msg`.
 + `--translateTime` (`-t`): Translate the epoch time value into a human readable
-in `UTC`. The default pattern is `'yyyy-mm-dd HH:MM:ss.l o'` time in UTC.
-IF want to translate to the local system's timezone, require `SYS:` prefix at 
-input format string. For a list of available pattern letters see the 
-[`dateformat` documentation](https://www.npmjs.com/package/dateformat#mask-options).
+date and time string in `UTC`. The default pattern is `'yyyy-mm-dd HH:MM:ss.l o'`.
+If you want to translate to the local system's timezone, then you must prefix the format 
+string with `SYS:`, e.g. `'SYS:yyyy-mm-dd HH:MM:ss'`. See [`dateformat` documentation](https://www.npmjs.com/package/dateformat#mask-options)
+for more available pattern letters.
 
 <a id="api"></a>
 ## API
--- a/Readme.md
+++ b/Readme.md
@@ -47,13 +47,13 @@ of properties. The list should be a comma separated list of properties Default: 
 error like objects. Default: `err,error`.
 + `--messageKey` (`-m`): Define the key that contains the main log message.
 Default: `msg`.
-+ `--systemTime` (`-n`): When translating the time to a human readable format,
++ `--systemZone` (`-s`): When translating the time to a human readable format,
 use the system timezone for displaying the time. See `--translateTime` for more
 information on the output format.
 + `--translateTime` (`-t`): Translate the epoch time value into a human readable
 UTC date and time string. The default format string is `'yyyy-mm-dd HH:MM:ss.l o'`.
-For a list of available patter letters see the 
-[`dateformat` documentation](https://www.npmjs.com/package/dateformat).
+For a list of available pattern letters see the 
+[`dateformat` documentation](https://www.npmjs.com/package/dateformat#mask-options).
 
 <a id="api"></a>
 ## API
@@ -69,7 +69,7 @@ in [CLI Arguments](#cliargs):
   errorLikeObjectKeys: ['err', 'error'], // --errorLikeObjectKeys
   errorProps: '', // --errorProps
   levelFirst: false, // --levelFirst
-  systemTime: false, // --systemTime
+  systemTime: false, // --systemZone
   messageKey: 'msg', // --messageKey
   translateTime: false // --translateTime
 }
--- a/Readme.md
+++ b/Readme.md
@@ -47,12 +47,10 @@ of properties. The list should be a comma separated list of properties Default: 
 error like objects. Default: `err,error`.
 + `--messageKey` (`-m`): Define the key that contains the main log message.
 Default: `msg`.
-+ `--systemZone` (`-s`): When translating the time to a human readable format,
-use the system timezone for displaying the time. See `--translateTime` for more
-information on the output format.
 + `--translateTime` (`-t`): Translate the epoch time value into a human readable
-UTC date and time string. The default format string is `'yyyy-mm-dd HH:MM:ss.l o'`.
-For a list of available pattern letters see the 
+in `UTC`. The default pattern is `'yyyy-mm-dd HH:MM:ss.l o'` time in UTC.
+IF want to translate to the local system's timezone, require `SYS:` prefix at 
+input format string. For a list of available pattern letters see the 
 [`dateformat` documentation](https://www.npmjs.com/package/dateformat#mask-options).
 
 <a id="api"></a>
@@ -69,7 +67,6 @@ in [CLI Arguments](#cliargs):
   errorLikeObjectKeys: ['err', 'error'], // --errorLikeObjectKeys
   errorProps: '', // --errorProps
   levelFirst: false, // --levelFirst
-  systemTime: false, // --systemZone
   messageKey: 'msg', // --messageKey
   translateTime: false // --translateTime
 }
--- a/Readme.md
+++ b/Readme.md
@@ -40,12 +40,6 @@ $ npm install -g pino-pretty
 + `--colorize` (`-c`): Adds terminal color escape sequences to the output.
 + `--crlf` (`-f`): Appends carriage return and line feed, instead of just a line
 feed, to the formatted log line.
-+ `--dateFormat` (`-d`): Sets the format string to apply when translating the date
-to human readable format (see: `--translateTime`). The default format string
-is `'yyyy-mm-dd HH:MM:ss.l o'`. For a list of available patter letters
-see the [`dateformat` documentation](https://www.npmjs.com/package/dateformat).
-When the value is anything other than the default value, `--translateTime` is
-implied.
 + `--errorProps` (`-e`): When formatting an error object, display this list
 of properties. The list should be a comma separated list of properties Default: `''`.
 + `--levelFirst` (`-l`): Display the log level name before the logged date and time.
@@ -53,10 +47,13 @@ of properties. The list should be a comma separated list of properties Default: 
 error like objects. Default: `err,error`.
 + `--messageKey` (`-m`): Define the key that contains the main log message.
 Default: `msg`.
-+ `--localTime` (`-n`): When translating the time to a human readable format,
-use the system timezone for displaying the time.
++ `--systemTime` (`-n`): When translating the time to a human readable format,
+use the system timezone for displaying the time. See `--translateTime` for more
+information on the output format.
 + `--translateTime` (`-t`): Translate the epoch time value into a human readable
-date and time string. See `--dateFormat` for information on the output format.
+UTC date and time string. The default format string is `'yyyy-mm-dd HH:MM:ss.l o'`.
+For a list of available patter letters see the 
+[`dateformat` documentation](https://www.npmjs.com/package/dateformat).
 
 <a id="api"></a>
 ## API
@@ -69,11 +66,10 @@ in [CLI Arguments](#cliargs):
 {
   colorize: false, // --colorize
   crlf: false, // --crlf
-  dateFormat: 'yyyy-mm-dd HH:MM:ss.l o', // --dateFormat
   errorLikeObjectKeys: ['err', 'error'], // --errorLikeObjectKeys
   errorProps: '', // --errorProps
   levelFirst: false, // --levelFirst
-  localTime: false, // --localTime
+  systemTime: false, // --systemTime
   messageKey: 'msg', // --messageKey
   translateTime: false // --translateTime
 }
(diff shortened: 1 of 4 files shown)


jsumners:
I'm not sure about this change. The current implementation is a direct copy of the `pino@4` implementation. My goal was to keep the transition as easy as possible. I agree that this is the more sane way of implementing the features.

What do you guys think @mcollina and @davidmarkclements? Given that this module is intended to be used with a new major of `pino`, I am willing to accept this PR.

index.js
```diff
@@ -56,10 +56,19 @@ module.exports = function prettyFactory (options) {
   const errorLikeObjectKeys = opts.errorLikeObjectKeys
   const errorProps = opts.errorProps.split(',')
 
-  if (opts.dateFormat.length > 0 && opts.dateFormat !== CONSTANTS.DATE_FORMAT) {
+  if (opts.localTime) {
+    if (opts.localTime.length > 0) {
+      opts.dateFormat = opts.localTime
```
jsumners: With this PR, `opts.dateFormat` is obsolete. It should be made into a local variable only and removed from the list of options.

victor0801x:
@jsumners document changed. and replace localTime by systemTime (this is more accurate)

pino does not document those options and usages !!
maybe affected people less than we though ?

Readme.md
```diff
@@ -40,23 +40,20 @@ $ npm install -g pino-pretty
...
 Default: `msg`.
-+ `--localTime` (`-n`): When translating the time to a human readable format,
-use the system timezone for displaying the time.
++ `--systemTime` (`-n`): When translating the time to a human readable format,
+use the system timezone for displaying the time. See `--translateTime` for more
+information on the output format.
 + `--translateTime` (`-t`): Translate the epoch time value into a human readable
-date and time string. See `--dateFormat` for information on the output format.
+UTC date and time string. The default format string is `'yyyy-mm-dd HH:MM:ss.l o'`.
+For a list of available patter letters see the
```
jsumners: Please fix my typo: "patter" => "pattern"

jsumners:
The lack of documentation around the date time formatting is a primary motivation for this module.

Readme.md
```diff
@@ -40,23 +40,20 @@ $ npm install -g pino-pretty
...
 + `--errorProps` (`-e`): When formatting an error object, display this list
 of properties. The list should be a comma separated list of properties Default: `''`.
 + `--levelFirst` (`-l`): Display the log level name before the logged date and time.
 + `--errorLikeObjectKeys` (`-k`): Define the log keys that are associated with
 error like objects. Default: `err,error`.
 + `--messageKey` (`-m`): Define the key that contains the main log message.
 Default: `msg`.
-+ `--localTime` (`-n`): When translating the time to a human readable format,
-use the system timezone for displaying the time.
++ `--systemTime` (`-n`): When translating the time to a human readable format,
```
davidmarkclements: do we have an `-s` flag? if not, why not `-s`?

davidmarkclements:
* it's a major version so breakage is acceptable
* this part is developer tooling, so even more so

I don't think `--systemTime` should be able the format, I think that should be a simple boolean modifier to `--translateTime` (and throw if not used in conjunction with translate time).

That way there's only one flag where you set the format.

davidmarkclements:
My thought was, only have one place where
formatting can be set - and -t sets the
date format (and if none supplied it uses default). So if -t is used as a
Boolean we're
saying apply human readable time

If -s is used we're saying output that time
as system time rather than utc - it's co-dependent

However just had a new thought, what if you
could specify utc or system time in the date format template itself?

jsumners:
@davidmarkclements you can. The only thing the current `--localTime` switch does is to not prepend `UTC:` to the default format string:

https://github.com/pinojs/pino-pretty/blob/78e000f7453f67a29f514f44d588fe9d170d5731/index.js#L41-L45

davidmarkclements:
So then -s is redundant - we can check for prepended timezone right? Is there support in the template format for a time zone inherited from the system?

jsumners:
The default is the system timezone. You have to opt-in to UTC. And those are your only options with the `dateformat` module.

davidmarkclements:
but we want utc as default.. 

what if we introduce a `SYS:` prefix to denote system time - which will *not* prefix `UTC:` otherwise `UTC` is prefixed?

davidmarkclements:
or we don't prefix UTC, and you opt in with your format

but the default includes UTC

jsumners:
I'm saying that `dateformat`, the module, by default formats `Date` objects according to the system timezone. `pino-pretty`, by default`, formats them according to UTC.

davidmarkclements:
yep

and we can either

* break default behaviour in the major and make all unprefixed time formats local time, *but* prefix UTC: to the default time format
* **OR** keep default behaviour and introduce an alternative prefix (`SYS:`) which can be used to *opt out* of automatically prefixing `UTC:`

davidmarkclements:
agreed I think we need to keep UTC the default – and it would be by dint the default
format having the UTC prefix

but for anything else you would need to specify UTC when supplying the format 

I think this is fine, I think it's good to allow for system time but not give it first class
citizenship via a flag – obviously this would be breaking but yeah - major, and devtool - not massively problematic. 

But I'm not super passionate either way, if we feel it's better to keep `-s` then cool

davidmarkclements:
OR (sorry to harp on) - would it be simpler: 

1. default to outputting human readable time in UTC
1. accept a format string to alter the output of that time
1. document the need for UTC prefix
1. leave the rest up to the user

We can even include the timeformat without the UTC prefix in the usage, 
as an example of system time

This would be a very trivial in terms of code, it's simply case of presenting it in docs and usage text

jsumners:
I was going for enduser ease of use. It's easier to:

+ `-t 'yyyy-mm-dd HH:MM'` (UTC)
+ `-t 'SYS:standard` (local with 'yyyy-mm-dd HH:MM:ss.SSS o')
+ `-t 'SYS:yyyy-mm-dd HH:MM`

davidmarkclements:
* do we recommend not using UTC to the extent that we want to go extra efforts to make it easy?
* on a certain level SYS:standard is just another way of saying `--systemTime` but just more unusual and more complex to implement

I think the crux here is: how important is system time for the average user, should we priotize it?
If we think it passes a threshold of importance where there are significant gains to supporting it, then I would turncoat and say stick with the `-s` flag, even allow for formatting in it.

jsumners:
@davidmarkclements: @victor0801x wanted it bad enough to put up with us on https://github.com/pinojs/pino/pull/280

victor0801x:
lastest change:
  - `-t` UTC in default pattern
  - `-t formatString` UTC in formatString
  - `-t 'SYS:standard'` local timezone in default pattern
  - `-t 'SYS:formatString'` local timezone in formatString
IMO, 
for Ops: local system zone is more easy and straight to use
for progs: utc is more consistent

Readme.md
```diff
@@ -40,23 +40,18 @@ $ npm install -g pino-pretty
...
 + `--levelFirst` (`-l`): Display the log level name before the logged date and time.
 + `--errorLikeObjectKeys` (`-k`): Define the log keys that are associated with
 error like objects. Default: `err,error`.
 + `--messageKey` (`-m`): Define the key that contains the main log message.
 Default: `msg`.
-+ `--localTime` (`-n`): When translating the time to a human readable format,
-use the system timezone for displaying the time.
 + `--translateTime` (`-t`): Translate the epoch time value into a human readable
-date and time string. See `--dateFormat` for information on the output format.
+in `UTC`. The default pattern is `'yyyy-mm-dd HH:MM:ss.l o'` time in UTC.
```
jsumners: "date and time string" should be inserted between "readable" and "in `UTC`."

jsumners: Once that is added, strike "time in UTC" from the second sentence.

Readme.md
```diff
@@ -40,23 +40,18 @@ $ npm install -g pino-pretty
...
+IF want to translate to the local system's timezone, require `SYS:` prefix at
```
jsumners: Replace this sentence with: "If you want to translate local system's timezone then you must prefix the format string with `SYS:`, e.g. `SYS:yyyy-mm-dd HH:MM:ss`."

bin.js
```diff
@@ -11,18 +11,17 @@ const CONSTANTS = require('./lib/constants')
...
   .option(['c', 'colorize'], 'Force adding color sequences to the output')
   .option(['f', 'crlf'], 'Append CRLF instead of LF to formatted lines')
-  .option(['d', 'dateFormat'], 'A format string to govern display of dates. When not set to the default value, `--translateTime` is implied', CONSTANTS.DATE_FORMAT)
   .option(['e', 'errorProps'], 'Comma separated list of properties on error objects to show (`*` for all properties)', '')
   .option(['l', 'levelFirst'], 'Display the log level as the first output field')
   .option(['k', 'errorLikeObjectKeys'], 'Define which keys contain error objects (`-k err,error`)', 'err,error')
   .option(['m', 'messageKey'], 'Highlight the message under the specified key', CONSTANTS.MESSAGE_KEY)
-  .option(['n', 'localTime'], 'Display timestamps according to system timezone')
-  .option(['t', 'translateTime'], 'Convert Epoch timestamps to ISO format')
+  .option(['t', 'translateTime'], 'Display epoch timestamps as UTC ISO format, also accept a formatString to alter the output (default ISO 8601)')
```
jsumners: "Display epoch timestamps as UTC ISO format or according to an optional format string (default ISO 8601)"

bin.js
```diff
@@ -11,18 +11,17 @@ const CONSTANTS = require('./lib/constants')
...
 
 args
   .example('cat log | pino-pretty', 'To prettify logs, simply pipe a log file through')
   .example('cat log | pino-pretty -m fooMessage', 'To highlight a string at a key other than \'msg\', use')
   .example('cat log | pino-pretty -t', 'To convert Epoch timestamps to ISO timestamps use the -t option')
+  .example('cat log | pino-pretty -t "SYS:formatString"', 'To convert Epoch timestamps to local timezone format use the -t option with SYS:formatString')
```
jsumners: "cat log | pino-pretty -t "SYS:yyyy-mm-dd HH:MM:ss"

index.js
```diff
@@ -38,10 +36,18 @@ function isPinoLog (log) {
...
-function formatTime (epoch, formatString, localTime) {
+function formatTime (epoch, translateTime) {
   const instant = new Date(epoch)
-  if (localTime) return dateformat(instant, formatString)
-  return dateformat(instant, 'UTC:' + formatString)
+  if (translateTime === true) {
+    return dateformat(instant, 'UTC:' + CONSTANTS.DATE_FORMAT)
+  } else {
+    const upperFormat = translateTime.toUpperCase()
+    return (upperFormat.indexOf('SYS:') === -1)
```
jsumners: `(upperFormat.startsWith('SYS:') === false)`

Readme.md
victor0801x: this ?

+ `--translateTime` (`-t`): Translate the epoch time value into a human readable
date and time string in `UTC`. The default pattern is `'yyyy-mm-dd HH:MM:ss.l o'`.
IF you want to translate local system's timezone, then you must prefix the format 
string with `SYS:`, e.g. `'SYS:yyyy-mm-dd HH:MM:ss'`. For a list of available patter
letters see the [`dateformat` documentation](https://www.npmjs.com/package/dateformat#mask-options).

```
+ `--translateTime` (`-t`): Translate the epoch time value into a human readable
date and time string in `UTC`. The default pattern is `'yyyy-mm-dd HH:MM:ss.l o'`.
IF you want to translate local system's timezone, then you must prefix the format 
string with `SYS:`, e.g. `'SYS:yyyy-mm-dd HH:MM:ss'`. For a list of available patter
letters see the [`dateformat` documentation](https://www.npmjs.com/package/dateformat#mask-options).
```

jsumners: Yes, but fix "IF" in the third sentence to "If".

victor0801x: sorry, which one?

+ `--translateTime` (`-t`): Translate the epoch time value into a human readable
date and time string in `UTC`. The default pattern is `'yyyy-mm-dd HH:MM:ss.l o'`.
If you want to translate local system's timezone, then you must prefix the format 
string with `SYS:`, e.g. `'SYS:yyyy-mm-dd HH:MM:ss'`. See [`dateformat` documentation](https://www.npmjs.com/package/dateformat#mask-options)
for more available pattern letters.

+ `--translateTime` (`-t`): Translate the epoch time value into a human readable
date and time string in `UTC`. The default pattern is `'yyyy-mm-dd HH:MM:ss.l o'`.
If you want to translate local system's timezone, then you must prefix the format 
string with `SYS:`, e.g. `'SYS:yyyy-mm-dd HH:MM:ss'`. For a list of available patten
letters see the [`dateformat` documentation](https://www.npmjs.com/package/dateformat#mask-options).

jsumners:
Given that this is a breaking change, I am going to wait to merge it. As a result of https://github.com/pinojs/pino/pull/386 this module will need to be updated with another breaking change. So I will implement both at the same time.

@davidmarkclements are you satisfied with the current state of this PR?

jsumners:
I'm merging this into `next-major`. @davidmarkclements if you want to change it before that hits release (same time as pino@5), we can get it sorted.

PR #11, opened 2018-04-14
Deal with prettifier function returning `undefined`

jsumners:
To support https://github.com/pinojs/pino/issues/395


jsumners:
@mcollina I have added the check in `asMetaWrapper`. However, I'm leaving it untested for now because it's too complicated to write a test with the current state of `pino`. Since we will be removing `asMetaWrapper` from here, it's just not worth the effort.

PR #25, opened 2018-08-16
(faster) kleur instead of chalk.

aichholzer:


mcollina:
Thanks @aichholzer, but this is a development tool, and not something to use in production. As such, we don't have benchmarks in place to measure any performance gains that this would have. I'm keen on keeping in Chalk, mainly because I never had any problems with it.

@jsumners what do you think?

davidmarkclements:
I think we need to put stronger constraints to prevent prettification in production 

ideas:

* check for NODE_ENV production and crash 
* detect if in a container, and crash 
* additional warning messages when process starts using pino-pretty in-process
* use npm deprecate as a hack to add a warning upon installation
* additional documentation


I also think we should reject PR's that aim to increase the speed of this library (unless it's too slow for development, which, I don't think it is) - because the open reason for doing that is when thinking of pino-pretty for in-process production logging

jsumners:
I think our documented warnings are sufficient. I do not think we should be actively trying to hinder someone from doing something despite our advice. We merely just refuse to support the usage we have recommended against.

aichholzer:
Gentlemen, good day!

@mcollina -You are welcome and agree but I am quite religious when it comes to the code I use and write, and the fact that something is not supposed to be used in production does not mean it should not be good or fast.

@davidmarkclements -I disagree, completely. You write open source software for people to use, in any way they please. If I want to put your tool in my production software then that it my problem and that is what your disclaimer is for. Writing open source does not mean you need to be the nanny of everyone using it. My "open" reason for this PR is not along the lines of what you thought it was, it was for what I said above. And again, in any case, I am free to decide whatever I put and run on my production server.

@jsumners -I agree. A note about the intended purpose of your product should suffice; the rest is up to the user.

Having that said, I was only trying to help. Feel free to reject my PR.

jsumners:
@mcollina can you cast a deciding vote? I think it is clear that Dave is opposed.

aichholzer:
@jsumners -Offense? None taken, just surprised by the fact. Also, I do not believe I made any assumptions about anyone's reason to do "open source"; all I implied was that if you make your code free, open source, then let people decide how to use it. I still maintain that; "Free code, free choice"

In retrospect, however, perhaps I should have taken an offense by the fact that somebody who does not know me at all, or a single detail about me for that matter, makes an absolute statement about why I am doing something, quote:

> because the open reason for doing that is when thinking of pino-pretty for in-process production logging

Having a discussion is one thing -and I appreciate a good one- being absolute about something while being ignorant of the facts or reasons, is a complete different thing. I do not partake in the latter.

davidmarkclements:
@aicholzer you're absolutely free to use how you please, and contributions are always welcome. Whether we accept or not depends on the guiding philosophies of the project. For this project, production performance is paramount. I've reread my comment, and I think my tone could be better - for that I apologise. Thanks for the PR.

Issue #54, opened 2019-03-02
Ability to format log string before message is appended


rayvincent-bsd:
This tool is exactly what I'm after, but I was hoping that it would be a little easier to format the log that's output before the message is appended.  I've built a solution and am hoping that it can be adopted into the project.  I see room here to provide configuration logic that can allow specifying what keys and delimiters should be rendered.

Rather than make a separate transport, I took a stab at adding a feature that would provide this the flexibility I was looking for.  This issue is going to be followed by a PR for review.

My specific use case, was to bind the module and method to each log entry and have them show in a format that's easy to digest.  I found a lot of modules that wrap pino and leverage CallSite, but that doesn't help with my minified code.  So I went with this approach:

```javascript
// logger/LoggerFactory.ts
/**
 * @module logger
 */
import * as pino from 'pino';
import * as pkg from '../package.json';

process.env.LOG_LEVEL = (process.env.LOG_LEVEL)? process.env.LOG_LEVEL : 'info';

/**
 * Factory class which will create a pino logger for each class to use.
 */
export default class LoggerFactory {
    /**
     * Pino Logger instance to be shared by all modules in this application
     * 
     * log level defaults to `info` and can be overridden with `$LOG_LEVEL` env variable
     */
    private static logger = pino({
        level: process.env.LOG_LEVEL,
        redact: { paths: ['hostname'], remove: true },
        prettyPrint: {
            translateTime: true,
            format: [
                { delimiter: '[', requiresAll: ['time'] },
                { key: 'time' },
                { delimiter: '] ', requiresAll: ['time'] },
                { key: 'level' },
                { delimiter: ' (' },
                { key: 'pid' },
                { delimiter: ') [' },
                { key: 'app' },
                { delimiter: '] [' },
                { key: 'class' },
                { delimiter: '] ' },
                { delimiter: '[', requiresAll: ['fn'] },
                { key: 'fn' },
                { delimiter: '] ', requiresAll: ['fn'] }
            ]
        }
    });

    /**
     * The module/class path where the logger will be used
     * 
     * @param classPath
     */
    public static getLogger(classPath:string):pino.Logger {
        return LoggerFactory.logger.child({
            app: pkg.logCode,
            class: classPath
        });
    }
}
```

```javascript
// api/mqtt/Client.js
/**
 * @module api/mqtt
 */
import LoggerFactory from '../../logger/LoggerFactory'
const logger = LoggerFactory.getLogger('api/http/Client');

export default class Client {
[...]

PR #55, opened 2019-03-02
#54: Add ability to format log up to when message is appended

rayvincent-bsd:
The idea here is to provide a token based representation of what the log should look like with some logical operators to help determine if a delimiter should be shown or not.

As you can see here' the `format` option is what would provide this functionality:
```javascript
{
  colorize: chalk.supportsColor, // --colorize
  crlf: false, // --crlf
  errorLikeObjectKeys: ['err', 'error'], // --errorLikeObjectKeys
  errorProps: '', // --errorProps
  levelFirst: false, // --levelFirst, Ignored if `format` is provided
  messageKey: 'msg', // --messageKey
  translateTime: false, // --translateTime
  search: 'foo == `bar`', // --search
  format: [ // Configurable format of the log up to when the message is appended
      { delimiter: '[', requiresAll: ['time'] },
      { key: 'time' },
      { delimiter: '] ', requiresAll: ['time'] },
      { key: 'level' }
      { delimiter: ' (', requiresOne: ['name', 'pid', 'hostname'] },
      { key: 'name' },
      { delimiter: '/', requiresAll: ['name', 'pid'] },
      { key: 'pid' },
      { delimiter: ' on ', requiresAll: ['hostname'] },
      { key: 'hostname' },
      { delimiter: ')', requiresOne: ['name', 'pid', 'hostname'] },
      { delimiter: ': ' },
  ]
}
```


index.js
```diff
@@ -108,54 +108,82 @@ module.exports = function prettyFactory (options) {
...
+        } else {
+          line += token.delimiter
+        }
       }
-
-      line += ')'
-    }
-
-    line += ': '
+    })
```
mcollina: Can you move this formatting logic into its own file under `lib/`?

rayvincent-bsd:
Of course, that'll be my next move. Wanted to make sure this feature had legs first.

index.js
rayvincent-bsd: Sure thing, I'll move it over

rayvincent-bsd: So I started to move this logic to a module in ./lib, but I realized quickly that this effort isn't worth it in my opinion.  There are too many variables that would need to be shared between the two modules that make this more difficult to understand without refactoring most of the codebase and introducing more risk.  I'm sure that's why the original logic was written in index.js as well.  At the end of the day, I've replaced 38 lines ([L120](https://github.com/pinojs/pino-pretty/blob/master/index.js#L120) -L158) of code with 44 lines ([L152](https://github.com/pinojs/pino-pretty/pull/55/files#diff-168726dbe96b3ce427e7fedce31bb0bcR152)-L196).

rayvincent-bsd:
I've added documentation, unit tests and cleaned up a few things.  Let me know if this needs anything more.

The only thing I think would make this more appealing would be to add a cli flag to read the options from a json file.  Of course, all previous options could easily be configured via command line, but with these new options, it would make more sense to have that read from a file.  That way you can take advantage of this formatting when piping the logs from another process.

Readme.md
```diff
@@ -117,16 +117,327 @@ with keys corresponding to the options described in [CLI Arguments](#cliargs):
...
+| hostname 	| The hostname of server where the application is running |
+| v        	| The pino log entry version 	|
+
+You can also bind as many properties to a log message or a logger via various apis.  This formatter will allow you to specify how each of the available keys will be logged in the first line, as well as how delimiters should be displayed depending on what log keys are available.
+
+##### key
+By specifiying a key entry in your format object, you are telling pino-pretty to attempt to output the value of this key in the log entry if available.
+
+```javascript
+interface KeyToken {
```
jsumners: What is an `interface`?  It looks to be in the future reserved words list, but isn't a feature of JavaScript -- https://www.ecma-international.org/ecma-262/9.0/index.html#sec-future-reserved-words

jsumners:
This seems to drastically complicate an already difficult to maintain module. I think we did a fair amount of work in Pino v5 to make it easy for people to write new prettifiers. I wonder if it wouldn't be better to do that instead? This module could even be forked as the start of such a module.

@davidmarkclements do you have any opinions here?

Readme.md
rayvincent-bsd: That was me just trying to explain what each of the available tokens can have.  I was referencing TypeScripts syntax for those familiar with it.  An interface is an age old OOP construct.

rayvincent-bsd:
> This seems to drastically complicate an already difficult to maintain module. I think we did a fair amount of work in Pino v5 to make it easy for people to write new prettifiers. I wonder if it wouldn't be better to do that instead? This module could even be forked as the start of such a module.
> 
> @davidmarkclements do you have any opinions here?

@jsumners, I took a look at the documentation for pino v5 and it pointed me directly to this project as reference for how to write a prettifier.  I personally feel that if you're able to expose this kind of configuration to the pino-pretty module, then it will give flexibility to anybody for almost any format that they would like to output.  The first line of output doesn't need that much logic to determine what to show so you likely won't find very complicated configurations.  And for those that need it, they can write their own.

I think this gives the extensibility to this prettifier so that whoever uses it doesn't need to fork this and write more code as long as you feel it's strong.

If the concern is that this module has evolved into spaghetti code and is difficult to maintain, then I would agree and it would benefit from being overhauled with modern OOP JS design patterns.  But that seems like an involved overhaul that may not be worth the effort/risk for as well as this performs now.

Just my two cents!

Readme.md
jsumners: I was playing dumb. The docs should be readable by people familiar with JavaScript and no other language since this is a JavaScript based project. I haven't had the time or fortitude to read through all of this to make a better suggestion other than "this isn't it."

meotimdihia:
I think module need be able to format output message. From my current CLI workers, I can't use Pino because it outputs too much useless info: hostname, process Id, ...

davidmarkclements:
I agree, this should be a separate module - and if it works out that it's objectively better (or at least better in some circumstances) than pino pretty we'd be happy to advertise and/or include it in the org. Closing

PR #56, opened 2019-03-04
Add ability to load run configuration from json file

rayvincent-bsd:
This will allow specifying run configurations loaded from `./.pino-prettyrc.json` by default or specifying a path via a new --options-file cli param.

### DIFF
--- a/test/cli-rc.test.js
+++ b/test/cli-rc.test.js
@@ -1,0 +1,48 @@
+'use strict'
+
+const path = require('path')
+const spawn = require('child_process').spawn
+const test = require('tap').test
+
+const bin = require.resolve(path.join(__dirname, '..', 'bin.js'))
+const logLine = '{"level":30,"time":1522431328992,"msg":"hello world","pid":42,"hostname":"foo","v":1}\n'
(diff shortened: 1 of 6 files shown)


test/cli-rc.test.js
```diff
@@ -0,0 +1,48 @@
...
+const test = require('tap').test
+
+const bin = require.resolve(path.join(__dirname, '..', 'bin.js'))
+const logLine = '{"level":30,"time":1522431328992,"msg":"hello world","pid":42,"hostname":"foo","v":1}\n'
+const rcFile = path.join(process.cwd(), '.pino-prettyrc.json')
+
+test('cli', (t) => {
+  t.afterEach((done) => {
+    // Update .pino-prettyrc.json with empty options
+    require('fs').writeFileSync(rcFile, JSON.stringify({}, null, 4))
```
mcollina: This is not really reliable. You should be creating folders and move the current process there instead to test this. `process.chdir` shoud do the trick.

rayvincent-bsd: Great point.  That's a much better idea.  I went ahead and made the changes so it creates a tmp directory and changes to it before each test

PR #58, opened 2019-03-30
[WIP] Plain ndjson

jsumners:
This PR will refactor `pino-pretty` to be a generic ndjson prettifier. It will recognize various properties that are common in ndjson _log lines_, and handle those properties in a sane way, but it will no longer skip JSON lines it does not recognize. Instead, it will apply a basic format to them.

TODO:

- [ ] ~Figure out what format bog standard JSON lines should be presented in~ (I do not have time for this)
- [x] Clean up the mess that is object and errors formatting
- [x] Write some tests for the new utility functions
- [x] Docs!

For the formatting of standard JSON lines I'm thinking:

```json
{"foo":"bar"}
```
=>
```js
{
  foo: "bar"
}
```

and

```json
["foo","bar"]
```
=>
```js
[
  "foo",
  "bar"
]
```

Opinions @pinojs/core?


jsumners:
PS: PRs to this branch are welcome if you see something you want to tackle.

jsumners:
I intend to.

jsumners:
FYI: stack formatting is a nightmare.

jsumners:
I think this is as done as I have time for. It would be nice to get a robust set of sample data to feed through it to at least generate a screenshot for the readme, but that can be a "good first issue" or I will come back to it later. We might as well wait for Node 12 to be released before closing this out since it is so close and this PR drops Node 6.

In the meantime, if anyone can think of how to cover https://coveralls.io/builds/22789235/source?filename=bin.js#L45 in the tests, a PR is welcome. The only other block Coveralls thinks is not covered is most definitely covered as I have stepped through it via the test suite.

It would also be fantastic if someone with more patience and time for formatting would come through and make the output even prettier. But it's good enough as-is.

PR #59, opened 2019-04-11
Added tests

ovhemert:
Added some tests for `prettifyErrorLog` and `prettifyObject`.

Checks functionalities of the specified functions and restores code coverage.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -254,7 +254,7 @@ function prettifyObject ({ input, ident = '    ', eol = '\n', skipKeys = [], err
           result += line
         }
       }
-    } else if (keysToIgnore.includes(keyName) === false) {
+    } else {
       result += `${ident}${keyName}: ${joinedLines}${eol}`
     }
   }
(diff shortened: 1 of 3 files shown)


lib/utils.js
```diff
@@ -254,7 +254,7 @@ function prettifyObject ({ input, ident = '    ', eol = '\n', skipKeys = [], err
           result += line
         }
       }
-    } else if (keysToIgnore.includes(keyName) === false) {
+    } else {
```
jsumners: Why this change?

ovhemert: In the coverage report I noticed that the condition was always hit with the same result. I wanted to write a test that made sure that this line was also hit with a false condition.

Eventually, the comparison cannot resolve to another result, since keysToIterate will never have a key that is in the keysToIgnore. They were already filtered out in line 228:

`const keysToIterate = objectKeys.filter(k => keysToIgnore.includes(k) === false)`

Thus, the `else if` will always be hit, and can be an `else` instead.

jsumners: 👍

PR #67, opened 2019-05-24
Support custom timestamp property

### LINKED ISSUE #66: Option to customize timestamp field name
There is already an option for customizing `message` field name, would be great to have same one for timestamp. 

Reason for this -> we'd like to be able to store it as `@timestamp` for convenience of consumption from Kibana, but currently pino is not aware that this is a timestamp, so during pretty print it displays [undefined] as time.

kibertoad:
fixes #66


kibertoad:
Aporeciate the quick review and release!

jsumners:
https://github.com/pinojs/pino-pretty/releases/tag/v3.1.0

Issue #69, opened 2019-06-18
Crash when logging integer


Tiim:
When i directly log a number with console.log without using pino then peno-pretty crashes with the following error message:

```js
/code/node_modules/pino-pretty/lib/utils.js:178
   if ('level' in log === false) return undefined
               ^

 TypeError: Cannot use 'in' operator to search for 'level' in 2
     at prettifyLevel (/code/node_modules/pino-pretty/lib/utils.js:178:15)
     at pretty (/code/node_modules/pino-pretty/index.js:85:29)
     at Transform.transform [as _transform] (/code/node_modules/pino-pretty/bin.js:38:18)
     at Transform._read (/code/node_modules/pino-pretty/node_modules/readable-stream/lib/_stream_transform.js:177:10)
     at Transform._write (/code/node_modules/pino-pretty/node_modules/readable-stream/lib/_stream_transform.js:164:83)
     at doWrite (/code/node_modules/pino-pretty/node_modules/readable-stream/lib/_stream_writable.js:405:139)
     at writeOrBuffer (/code/node_modules/pino-pretty/node_modules/readable-stream/lib/_stream_writable.js:394:5)
     at Transform.Writable.write (/code/node_modules/pino-pretty/node_modules/readable-stream/lib/_stream_writable.js:303:11)
     at Transform.ondata (/code/node_modules/split2/node_modules/readable-stream/lib/_stream_readable.js:662:20)
     at Transform.emit (events.js:196:13)
```

The following snipped in my code (fastify server) will produde this crash:
```js
console.log(index); // index = 2
```

jsumners:
Looks like a `Number.isFinite(log)` check should be added at https://github.com/pinojs/pino-pretty/blob/a0631872bd65a01faf14c850348fb3b575ffff2e/index.js#L68

Care to submit a PR?

Tiim:
I would love to, but I don't have time at the moment.

Tiim:
Thanks #70 seems to fixed it 👍

PR #70, opened 2019-06-18
add check for numbers

ovhemert:
Checks for number input and prevents crashing as reported in issue #69

Example: `node -e 'console.log(2)' | node ./bin.js`

### DIFF
--- a/test/basic.test.js
+++ b/test/basic.test.js
@@ -349,6 +349,20 @@ test('basic prettifier tests', (t) => {
     t.end()
   })
 
+  t.test('handles numbers', (t) => {
+    const pretty = prettyFactory()
+    let formatted = pretty(2)
+    t.is(formatted, '2\n')
+
+    formatted = pretty(-2)
+    t.is(formatted, '-2\n')
+
+    formatted = pretty(0.2)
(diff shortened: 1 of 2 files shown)


test/basic.test.js
```diff
@@ -349,6 +349,20 @@ test('basic prettifier tests', (t) => {
...
+    const pretty = prettyFactory()
+    let formatted = pretty(2)
+    t.is(formatted, '2\n')
+
+    formatted = pretty(-2)
+    t.is(formatted, '-2\n')
+
+    formatted = pretty(0.2)
+    t.is(formatted, '0.2\n')
+
```
davidmarkclements: can you add a test for `Infinity` and `NaN` (since typeof NaN is number)

davidmarkclements: I predict these will fail

ovhemert: here you go

jsumners: I predict they will not parse as JSON and thus will not reach this check at all https://tools.ietf.org/html/rfc7159#section-3

davidmarkclements: you predict right 😆

Issue #62, opened 2019-04-23
Log level incorrectly shows up as USERLVL when `useLevelLabels: true`


tonyxiao:
I created pino with option `useLevelLabels: true`. When formatting via pino-pretty those log lines show up as 
```
[2019-04-23 13:56:29.560 +0000] USERLVL (runScript): Request did succeed
```
The log line itself cpontains
```
{
  "level": "info",
  "time": 1556027789560,
  "name": "runScript",
  "msg": "Request did succeed",
  "v": 1
}
```

Given that `useLevelLabels` is a feature built into pino, pino-pretty should be able to understand their corresponding string based levels.

jsumners:
1. That log line looks like it should be showing as `INFO`. Does your sample have a typo or did you find a bug?
2. It would be extraordinarily complex to make `pino-pretty` recognize any user defined levels in a `pino` instance. Pino custom levels are mutable and `pino-pretty` merely reads the stream `pino` writes to.

tonyxiao:
I'm pretty certain it's a bug, see

```bash
$ echo '{ "level": "info", "time": 1556027789560, "name": "runScript", "msg": "Request did succeed", "v": 1 }' | ./node_modules/.bin/pino-pretty
[1556027789560] USERLVL (runScript): Request did succeed

$ cat yarn.lock | grep pino-pretty
pino-pretty@^2.6.0:
  resolved "https://registry.yarnpkg.com/pino-pretty/-/pino-pretty-2.6.0.tgz#05e69e5d458e8b304936e0097045c38c382e7c27"
```

tonyxiao:
> It would be extraordinarily complex to make pino-pretty recognize any user defined levels in a pino instance. Pino custom levels are mutable and pino-pretty merely reads the stream pino writes to.
Outside of the standard ones defined by pino, couldn't pino-pretty simply output the level as-is rather than displaying "USERLVL"?

jsumners:
Ah, it is not a bug. `pino-pretty` 2.x does not support `"level": "info"`. It only support levels as integers. So it's a conflict between `pino` configuration and what `pino-pretty@2` supports. `pino-pretty@next` has you covered in this regard.

As for parroting the level, as hinted, the level can _either_ be a string or a number. It would also be difficult to maintain the alignment for level labels. That being said, feel free to submit a PR against my `plain-ndjson` branch if you would like to see this behavior changed. It's going to be at least Sunday before I can close out that branch.

[...]

PR #75, opened 2019-08-05
modular log builder

noamkfir:
This PR essentially refactors the `pretty` function to make it extensible.

The main reason for doing this is that it makes it possible to use and extend the built-in `pino-pretty` prettifier instead of creating a whole new prettifier from scratch, which is not as simple a task as it should be and also not very well documented. It's also usually a wasteful task because most custom prettifiers probably only need to make small modifications rather than completely starting from scratch.

The PR also includes new tests and documentation.

Two extension points were added, based on the existing code: log parsers, which process the ndjson; and line builders, which construct the strings that will eventually comprise the formatted line.


Readme.md
```diff
@@ -129,13 +129,115 @@ with keys corresponding to the options described in [CLI Arguments](#cliargs):
...
+
+Line builders are simple functions that prepare the final formatted line and are executed
+after the log parsers. `pino-pretty` uses a number of built-in line builders.
+
+The `lineBuilders` option accepts an array of functions, which are appended to the
+built-in line builder functions and therefore executed immediately after them. This
+provides an opportunity to further modify the formatted log output by supplying
+custom line builder functions.
+
+The log parser function reeives two parameters:`lineParts` and `context`.
```
jsumners:
```suggestion
The log parser function receives two parameters:`lineParts` and `context`.
```

lib/line-builders.js
```diff
@@ -0,0 +1,99 @@
+'use strict'
+
+const {
+  prettifyErrorLog,
+  prettifyObject
+} = require('./utils')
+
+const lineBuilders = [
+  (lineParts, { prettified, opts }) => {
```
jsumners: I'm not fond of this big array of anonymous functions. As written, it is not clear what each function is meant to do. Even so much as a proper function declaration with a name can make things much clearer. Even better would be a jsdoc block preceding each function. Look at `lib/utils.js` for a nice example.

lib/log-parsers.js
```diff
@@ -0,0 +1,62 @@
...
+const jsonParser = input => {
+  try {
+    return { value: bourne.parse(input, { protoAction: 'remove' }) }
+  } catch (err) {
+    return { err }
+  }
+}
+
+const logParsers = [
+  (input, { EOL }) => {
```
jsumners: Again, this is just a big blob of anonymous functions without any context. It will be very difficult to maintain.

index.js
```diff
@@ -38,110 +28,54 @@ const defaultOptions = {
...
 
-    if (!opts.levelFirst && prettifiedLevel) {
-      if (line.length > 0) {
-        line = `${line} ${prettifiedLevel}`
-      } else {
-        line = prettifiedLevel
+  function pretty (inputData) {
+    let nextInput = inputData
+
+    for (const logParser of logParsers) {
```
jsumners: A major concern I have with the rewrite of this file is the introduction of a lot of iteration. You iterate over "log parsers" which in turn iterates over "line builders". I bet if we had some benchmarks, and that's a failing on our part, we would see a significant difference in processing time with this method over the very linear method currently on `master`. Prettification is expensive.

index.js
```diff
@@ -38,110 +28,54 @@ const defaultOptions = {
...
+    EOL: opts.crlf ? '\r\n' : '\n',
+    IDENT: '    '
+  }
 
-    const prettifiedLevel = prettifyLevel({ log, colorizer })
-    const prettifiedMessage = prettifyMessage({ log, messageKey, colorizer })
-    const prettifiedMetadata = prettifyMetadata({ log })
-    const prettifiedTime = prettifyTime({ log, translateFormat: opts.translateTime, timestampKey })
+  if (opts.logParsers) {
+    logParsers.push(...opts.logParsers)
```
kuba-kubula: _Sorry to jump in into an interesting PR._

This line would modify the global `logParsers` exported from the `require('./lib/log-parsers')` line, so adding a new instance of `pino` with `prettyPrint` options would use the previously modified (pushed to) `logParsers` too, which is a conflict to me.

You should create local copy of `logParsers`, not modify the global ones.

jsumners: Good catch.

Readme.md
noamkfir: Fixed, thanks.

lib/line-builders.js
noamkfir: That's a good point. I just pushed a pretty big update that significantly improves this by using functionally cohesive classes with parse and build methods that make them far more readable and maintainable.

lib/log-parsers.js
noamkfir: Agreed, again. The update I just pushed improves both parsers and builders.

index.js
noamkfir: It's definitely a valid concern. I found an interesting benchmark (https://jsperf.com/loops/285) that compares the performance of different kinds of loops and discovered, to my surprise, that the `for of` loop is pretty slow. I've changed the loops to classic `for` loops, which are either the fastest loops or among the fastest loops, though I'm not sure if the benchmark is relevant for node on the desktop.

Having a benchmark for pino-pretty would be useful. But even without it, it's important to remember that we're talking about the prettifier and not the logger itself. Performance is extremely important but I do think that mitigates it a bit.

Finally, one of the advantages of this approach is that the prettifier can be optimized very easily by simply not using the parsers and builders that are irrelevant for an app. For example, if you don't want to show the timestamp, don't use the time log processor, thus avoiding the parsing, prettification, and building costs associated with it.

index.js
noamkfir: Definitely a good catch! I've applied your suggestion.

noamkfir:
Following up on the feedback, particularly the future maintenance aspect, I had an idea on how to further improve things. The additional changes improve performance, make the code far more readable (I hope), and make the prettifier much easier to extend.

I've replaced "log parsers" and "line builders" with "log processors". A log processor can either parse a log entry, contribute to building the output, or both. The new abstraction unifies and streamlines parsers, builders, and the various prettification functions with a very simple log processing API.

One interesting side effect of this change is that the complete sequence of operations undertaken by the prettifier is much more transparent now: json, primitives, search, ignore, time, level, metadata, semicolon, message, eol, error, object. When extending the prettifier, built-in log processors can be used along with custom log processors, and unneeded log processors can be easily removed as well.

I've also updated the tests and documentation.

I know this is an even more "radical" change than the initial pull request, but it really is a significant improvement.

noamkfir:
> I'm not fundamentally opposed to this PR, but I definitely need some convincing. This radically changes the way the script operates and could be difficult to maintain if you disappear. I'd like to see a very clear example of what this PR affords over the current implementation. As an example, does this PR make it easier to implement `pino-colada`? Can you show something like that? It is not clear from the included documentation what is possible with this new API.

I've provided a few small examples in the readme. I'm not familiar with `pino-colada`, but I'll take a look.

To answer the broader question, I initially wanted to accomplish what I thought was a pretty simple task. I add a few properties to some child loggers and I wanted to be able to show them in bracketed segments that prefix the log message, without adding additional lines for the object properties. The additional lines take up space and are difficult to parse visually as the text scrolls down the screen. The existing options make it possible to "ignore" these keys completely. but that's not what I needed.

I later realized that I'm not interested in the metadata. Simply removing the metadata log processor from the sequence improves both performance and the logs themselves.

I didn't want to create a new prettifier from scratch because I do need to preserve the majority of the prettifier's behavior and I don't really want to maintain a separate prettifier.

Other possibilities afforded by this solution include providing different output for entries marked with different log levels (e.g., 'info' lines and 'debug' lines produce slightly different output), adding line numbering to the output, using different characters for line breaks and indentation, and more. I haven't tried, but it's conceivable that it can even be used to transform the output into some other text formats, such as CSV.

lib/LogProcessor.js
```diff
@@ -0,0 +1,4 @@
+class LogProcessor {
+}
```
mcollina: What is the role of this base class? I think it can be completely removed.

lib/log-processors.js
```diff
@@ -0,0 +1,96 @@
+const LogProcessor = require('./LogProcessor')
```
mcollina: Can you please make sure that there is ‘use strict’ at the top of each file?

test/lib/utils.public.test.js
```diff
@@ -184,72 +196,85 @@ tap.test('prettifyObject', t => {
...
 
 tap.test('prettifyTime', t => {
-  const { prettifyTime } = utils
-
   t.test('returns `undefined` if `time` or `timestamp` not in log', async t => {
-    const str = prettifyTime({ log: {} })
-    t.is(str, undefined)
+    const prettified = {}
+    prettifyTime({}, { prettified })
+    t.is(prettified.prettifiedTime, undefined)
```
mcollina: I might have missed why we are modifying an object instead of returning a string. Would you mind to articulate?

index.js
jsumners: > But even without it, it's important to remember that we're talking about the prettifier and not the logger itself. 

Agreed. In particular when running as a piped process. But it still a concern nonetheless.

Readme.md
```diff
@@ -129,13 +129,276 @@ with keys corresponding to the options described in [CLI Arguments](#cliargs):
...
+
+#### examples
+
+The following are a number of examples of parsers, builders, and combined log processors:
+
+* **parser** - Perform a transformation on the input message:
+
+  ```js
+  function (input) {
+    input.msg = input.msg.toLowerCase()
```
jsumners: I'm torn. I don't like functions that have side effects. But the immediately obvious alternative incurs even more overhead. But maybe, and this is just a thought for discussion, the signature could be `function parser ({logObject, context, state, collector}) {}`. `logObject` would be a frozen instance of the originally parsed log line and the `collector` could be modified and passed along to ultimately be compiled into the output string.

@mcollina @davidmarkclements what are your opinions?

Readme.md
```diff
@@ -129,13 +129,276 @@ with keys corresponding to the options described in [CLI Arguments](#cliargs):
...
+    state.stop() // short-circuit the log parsing process
+    return input
+  }
+  ```
+
+* **parser** - Prepare a prettified string:
+
+  ```js
+  {
+    parse(input, { colorizer, inlineSection ) {
```
jsumners:
```suggestion
    parse(input, { colorizer, inlineSection }) {
```

Readme.md
```diff
@@ -129,13 +129,276 @@ with keys corresponding to the options described in [CLI Arguments](#cliargs):
...
+          lineParts.push(' ')
+        }
+        lineParts.push(prettifiedTime)
+      }
+    }
+  }
+  ```
+
+* **parser** and **builder** - Parse a log entry and then use the result to build the
+  output line:
```
jsumners: This sentence makes it sound like the `build` function will be used _immediately_ after the provided `parse` function is used. Is that the case? Or is the `build` function used during the build process?

lib/LogProcessor.js
noamkfir: The base class is used in `log-processors.js` in the `createLogProcessor` factory function to quickly identify built-in log processors. I suppose the `instanceof` checks can be replaced with member checks (e.g., `definition.parse || definition.build` or something similar instead, but the current implementation seems more readable.

I actually considered exposing it so that consumers could directly inherit from it, but I was concerned about changing the API.

test/lib/utils.public.test.js
noamkfir: The parse process is essentially a pipeline, in which the structure of the input and output of each function in the pipeline is identical. So a string can't be returned directly because it would be fed into the next parser and we'd lose the log object.

There are exceptions to this. The `primitives` parser returns a primitive value, but then it short-circuits the log processor, bypassing additional parsers and builders.

The input and output of the `parse` function is the log entry which is eventually fed to the builders, though it may be modified by the parsers as necessary. It seemed to make more sense to "cache" the prettified values on the `context.prettified` object because it mirrors the behavior that already existed before my changes and it also makes the API clearer by not "dirtying" the log object.

lib/log-processors.js
noamkfir: Sorry about that. Done.

Readme.md
```diff
@@ -129,13 +129,276 @@ with keys corresponding to the options described in [CLI Arguments](#cliargs):
...
+parsers, or between the parse and build stages of a log processor that supports both
+stages.
+
+`pino-pretty` uses the following built-in log processors, in this order:
+
+* `json` (*parser*) - Parses the input `ndjson` string as an object, which is
+  subsequently passed to the remaining log processors.
+* `primitives` (*parser*) - If the parsed result of the `json` log processor is `null`,
+  a boolean value, or a number, the log processing sequence for the current line is
+  aborted (short-circuited) and the value is returned.
```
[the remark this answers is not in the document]
noamkfir: I'm not sure I understand. The original behavior was preserved, so it is aborted in this case too. But I've added a note about this to the readme.

Readme.md
```diff
@@ -129,13 +129,276 @@ with keys corresponding to the options described in [CLI Arguments](#cliargs):
...
+* `search` (*parser*) - If the `search` property is specified, filters out entries that
+  do not match the `search` value, based on the `jmespath` library.
```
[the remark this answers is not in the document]
noamkfir: We do, on line 76:
https://github.com/pinojs/pino-pretty/pull/75/files#diff-1e290ac8433d555bce009b162cb869d0R76

Readme.md
```diff
@@ -129,13 +129,276 @@ with keys corresponding to the options described in [CLI Arguments](#cliargs):
...
+  ```js
+  {
+    parse(input, context) {}
+    build(lineParts, context) {}
+  }
+  ```
+* **parser** function
+  ```js
+  (input, context) => {}
+  ```
```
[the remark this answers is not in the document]
noamkfir: Now that you mention it, I agree.

I left it in because it was really easy to, but it now occurs to me that there may be another reason to get rid of the parser function: it creates an inconsistency with the builders because matching support for a builder function doesn't exist and can't easily be added.

Unless someone can come up with a really good reason to leave it in, I'll take it out.

Readme.md
noamkfir: I have to say that I share your concern. While I was working on this, I toyed with using pure functions that return an object like this (pseudo-code):

```js
const result = {
  output: { ...input, modifiedProperty, newProperty },
  done: false || true, // the previous incarnation of state.stopped
}
```

The problem with this approach is that it would almost certainly incur a significant hit to performance associated with the rapid creation and destruction of so many objects and primitives. Also, this approach would only create a shallow clone. A deep clone would hurt performance even more.

So that leaves us with unavoidable side-effects. A mitigating factor though is that the pipeline goes in one direction and we don't use complicating features such as events, rendering the side-effects relatively safe.

One of the problems with my approach that all of you have (thankfully!) pointed out in one way or another is that it's not very clear where to apply the side-effects: modify the `input` object, add values to the `context.prettified` object, or modify the `context` object directly.

The `collector` might be able to solve that problem, but it if it doesn't preserve the original structure of the log object, it would force us to rethink the way the builders work.

Also, if we used the the `collector`, it should be the second argument, not the last. `context` and `state` aren't always necessary, the latter being especially rare, but the collector would be used in almost every parser.

Readme.md
noamkfir: No, you're right. The `build` function is used during the build stage. I've updated the Readme to clarify.

noamkfir:
> > I'm not fundamentally opposed to this PR, but I definitely need some convincing. This radically changes the way the script operates and could be difficult to maintain if you disappear. I'd like to see a very clear example of what this PR affords over the current implementation. As an example, does this PR make it easier to implement `pino-colada`? Can you show something like that? It is not clear from the included documentation what is possible with this new API.
> 
> I've provided a few small examples in the readme. I'm not familiar with `pino-colada`, but I'll take a look.

@jsumners I put together a quick example based on `pino-colada`:
https://github.com/noamkfir/pino-colada/tree/pino-pretty-example

1. Run `npm install`
2. Run `node example-pino-pretty.js`
3. Open `http://localhost:8080`
4. Observe `pino-colada`-style log output.

The complete implementation is in the `pino-pretty-example.js` and `pino-pretty-processors.js` files. I basically copied the formatter implementations from the other examples. I chose to use log processors that handle both the parsing and building for each element, though it can also technically be optimized somewhat by unifying the builders.

noamkfir:
> Good work! Would you mind to add more unit tests for all the processors?

Thanks! I don't mind adding more unit tests. I'll try to add some soon.

I will note though that although the overall changes are significant, most of them were made by a sequence of simple refactorings, so the existing unit tests do cover the log processors for the most part.

So I didn't move and rename all the unit tests to match the new structure, but the existing suite of unit tests actually made it possible to make all these changes safely.

noamkfir:
> First off, I really appreciate the time being invested to improve this module that most definitely needs help. And the work is excellent quality. That being said, I have some concerns.
> 
> I don't like the way the functional nature of the code base has been removed in favor of "classes" that depend upon side effects. This is a surefire way to complicate things and introduce bugs. I do not use, or even think it should have ever been added to the language, the `class` syntax. From what I have read in this PR, the classes here are not complicated, but it would still require knowledge of what that sugar is doing under the hood to properly maintain them when bugs are reported. But more than that, the side effects really bother me.
> 
> Look at the time prettifier tests. They went from (paraphrased):
> 
> ```js
> const result = prettifyTime(input)
> t.is(result, 'a string')
> ```
> 
> To:
> 
> ```js
> prettifyTime(input)
> t.is(input.prettifiedTime, 'a string')
> ```
> 
> I think this highlights a fundamental issue that can, and more than likely will, show up as a bug report in the future. I left a comment in my review of the documentation that could alleviate this to a satisfactory degree; let's discuss it.
> 
> My other issue is that instead of discussing things after the initial review an essentially new PR was pushed up. My original suggestion was basically to refactor the code provided into:
> 
> ```js
> module.exports = [foo, bar]
> 
> function foo () {}
> 
> function bar () {}
> ```
> 
> But now the conversation is having to switch into more difficult territory that we could have decided upon together. Regardless, I'm not blocking this PR. If @mcollina and @davidmarkclements are happy with it, then I will be okay. But if more work is to be done, hopefully we can discuss it first so that effort doesn't feel wasted or unwelcome.

@jsumners Sorry about the lack of discussion. Of course, you're right. However, inspiration struck and I had to see it through :).

In any case, I did my best to stick with the stylistic preferences of the project. My personal preferences are a bit different, so I apologize if some of that leaked through.

The use of classes is intentionally very minimal precisely for the reasons you mentioned. *Complex* class structures tend to do more harm than good, in my experience. The classes I added can be easily removed if you all prefer.

The side-effects are a compromise between proper pure functions and immutable data structures on the one hand, and performance and clarity on the other. I mentioned this in a reply to another comment here somewhere. This is definitely something we should talk more about.

One thing that could help, besides the `collector` option, is a suite of performance tests. I'm not sure this is something I can do properly right now for this project, though.

faustbrian:
@noamkfir how do you think could https://github.com/pinojs/pino-pretty/issues/62 be resolved with the changes of this PR? (see lower half of the issue conversation)

jsumners:
I'm closing this due to inactivity. If you wish to continue championing this PR, please reopen.

PR #76, opened 2019-08-12
Handle timestamp 0 in prettifyTime

Zirak:
When trying to format a log object like the following:

```json
{
  "level": 30,
  "time": 0,
  "name": "foobar",
  "msg": "I like cake",
  "v": 1
}
```

pino-pretty previously tried to format the timestamp `undefined` instead of
taking the `0` time. This commit looks not at the truthiness of the value, but
its existence.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -326,9 +326,18 @@ function prettifyObject ({
  * string.
  */
 function prettifyTime ({ log, timestampKey = TIMESTAMP_KEY, translateFormat = undefined }) {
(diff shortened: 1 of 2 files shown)


Zirak:
Sorry, I was optimistic and did this on the fly. When I get to a saner environment I'll look at how badly I broke things!

mcollina:
Would you mind adding unit tests for this?

Il giorno lun 12 ago 2019 alle 17:45 Zirak <notifications@github.com> ha
scritto:

Zirak:
Tested and corrected my kerfuffle, sorry about the extra noise.

PR #77, opened 2019-08-14
Add `module` as a prettified property

NBTX:
In the Pino documentation, it is mentioned that [child loggers](https://github.com/pinojs/pino/blob/master/docs/child-loggers.md) can add properties to log output.

We have child loggers for different modules, as is imagined in the documentation and we use the `module` property to identify it.

This PR changes the log output such that it will include `[module] <message>`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -8,6 +8,7 @@ const {
   isObject,
   prettifyErrorLog,
   prettifyLevel,
+  prettifyModule,
   prettifyMessage,
   prettifyMetadata,
   prettifyObject,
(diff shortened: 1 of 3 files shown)


jsumners:
I think it is better to settle #75 first. If that gets merged then this PR doesn't need to go into core. It can simply be a configuration option passed in.

NBTX:
Ahh. Okay in that case I'll wait for #75. If it's a no for that, I'll come back and write unit tests. Otherwise I guess I'll write a module to do this instead.

mcollina:
Closing for no activity, feel free to reopen if you are still interested.

PR #84, opened 2019-11-01
Load options from config file

vladimyr:
This is extension of #56. It allows loading prettifier options from sibling `pino-pretty.config.js`, `.pino-prettyrc` or `.pino-prettyrc.json` configuration file (listed by priority) by default or custom one specified by `--config` cli flag.


package.json
```diff
@@ -36,9 +36,12 @@
     "dateformat": "^3.0.3",
     "fast-safe-stringify": "^2.0.7",
     "jmespath": "^0.15.0",
+    "joycon": "^2.2.5",
     "pump": "^3.0.0",
     "readable-stream": "^3.4.0",
-    "split2": "^3.1.1"
+    "rimraf": "^2.6.3",
```
jsumners: Why is this a regular dependency?

bin.js
```diff
@@ -2,11 +2,34 @@
...
+  parseJSON,
+  files: [
+    'pino-pretty.config.js',
+    '.pino-prettyrc',
+    '.pino-prettyrc.json'
+  ],
+  stopDir: path.dirname(process.cwd())
+})
+joycon.addLoader({
+  test: /\.[^.]*rc$/,
```
jsumners: Will this match files other than `pino-pretty` specific files?

package.json
vladimyr: You are right, failed to spot it. Will change to dev dependency.

bin.js
vladimyr: Joycon is limited to filenames listed above so we are fine there. In case custom filename is used it is actually preferable to interpret all possible rc files as json because one could invoke `pino-pretty` using `--config path/to/.customrc`.

jsumners: I really don’t understand what this is doing. There isn’t any documentation for this option that I can find.

vladimyr: It is literally second example inside Joycon's usage section: https://github.com/egoist/joycon#usage

Joycon by default only recognizes filenames with `.js` & `.json` extensions. In order to instruct it to treat _dot_ rc files as json (same like babel does) you need to plug in custom loader: https://github.com/egoist/joycon#addloaderloader
Required properties are:
- `test` which is regex that filename has conform to
- `load` or `loadSync` function that reads and parses given file(path) and returns data

PR #86, opened 2019-11-04
Update readme badges

vladimyr:
This PR updates readme badges. For convenience reasons I included npm package version badge and also switched existing ones to https://badgen.net using it as universal badge provider.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -1,8 +1,9 @@
 <a id="intro"></a>
 # pino-pretty
 
-[![Build Status](https://travis-ci.org/pinojs/pino-pretty.svg?branch=master)](https://travis-ci.org/pinojs/pino-pretty)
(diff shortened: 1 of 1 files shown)


mcollina:
Thanks for your contribution @vladimyr. Would you like to explain the rationale behind this change? I think our badges are good enough.

vladimyr:
Main reason I proposed it is to add npm version badge which serves informative purpose. It is just shortcut for users to quickly see what is latest published version without doing separate npmjs.com search. It is impossible to say looking at the code is current version published yet or not.
In the same time I altered other badges to use community powered badgen.net proxy because shields.io & badgen.net became standardish approach to readme badges and their appeal is in fact that they list badge url patterns for bunch different badge services on single page meaning it is super easy to find what you need compared to 3rd party documentation of originating service.

PR #88, opened 2019-11-08
feat: add messageFormat option for msg

shape-ideas:
Add `messageFormat` for combining some useful log properties into `msg`. And combined properties can be ignored.

Since `pino-pretty` is recommended for development only,  may `messageFormat` is a flexible way for module, for example:

```js
// log
{
  module: 'repository',
  msg: 'foo'
}

// options
{
  messageFormat: '[{module}] - {msg}',
  ignore: 'module',
}
```

The result will be `[repository] - foo`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -30,6 +30,7 @@ const defaultOptions = {
   errorProps: '',
   levelFirst: false,
   messageKey: MESSAGE_KEY,
+  messageFormat: false,
   timestampKey: TIMESTAMP_KEY,
   translateTime: false,
   useMetadata: false,
@@ -42,6 +43,7 @@ module.exports = function prettyFactory (options) {
   const EOL = opts.crlf ? '\r\n' : '\n'
   const IDENT = '    '
   const messageKey = opts.messageKey
+  const messageFormat = opts.messageFormat
   const timestampKey = opts.timestampKey
   const errorLikeObjectKeys = opts.errorLikeObjectKeys
   const errorProps = opts.errorProps.split(',')
@@ -75,6 +77,8 @@ module.exports = function prettyFactory (options) {
       return
     }
 
+    const prettifiedMessage = prettifyMessage({ log, messageKey, colorizer, messageFormat })
+
     if (ignoreKeys) {
       log = Object.keys(log)
         .filter(key => !ignoreKeys.has(key))
@@ -85,7 +89,6 @@ module.exports = function prettyFactory (options) {
     }
 
     const prettifiedLevel = prettifyLevel({ log, colorizer })
-    const prettifiedMessage = prettifyMessage({ log, messageKey, colorizer })
     const prettifiedMetadata = prettifyMetadata({ log })
     const prettifiedTime = prettifyTime({ log, translateFormat: opts.translateTime, timestampKey })
 
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -186,6 +186,8 @@ function prettifyLevel ({ log, colorizer = defaultColorizer }) {
  * @param {object} input.log The log object with the message to colorize.
  * @param {string} [input.messageKey='msg'] The property of the `log` that is the
  * message to be prettified.
+ * @param {string} [input.messageFormat=false] An option defines the output of
+ * msg. Default: false
  * @param {function} [input.colorizer] A colorizer function that has a
  * `.message(str)` method attached to it. This function should return a colorized
  * string which will be the "prettified" message. Default: a no-op colorizer.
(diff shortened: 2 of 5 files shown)


lib/utils.js
```diff
@@ -194,7 +194,16 @@ function prettifyLevel ({ log, colorizer = defaultColorizer }) {
  * key is not a string, then `undefined` will be returned. Otherwise, a string
  * that is the prettified message.
  */
-function prettifyMessage ({ log, messageKey = MESSAGE_KEY, colorizer = defaultColorizer }) {
+function prettifyMessage ({ log, messageFormat, messageKey = MESSAGE_KEY, colorizer = defaultColorizer }) {
+  if (messageFormat) {
+    const message = messageFormat.replace(/{([^}]+)}/g, function (match, p1) {
```
mcollina: Can you add some escaping support and tests?
So that somebody could use `\{{level}\} - \{{pid}\}`.

mcollina:
Good work and good addition!

index.js
```diff
@@ -75,6 +77,8 @@ module.exports = function prettyFactory (options) {
       return
     }
 
+    const prettifiedMessage = prettifyMessage({ log, messageKey, colorizer, messageFormat })
```
jsumners: Why does this action happen before ignoring keys now?

lib/utils.js
jsumners: Also, please update the docblock to describe the new parameter. Is it a required parameter? If not, what is the default?

jsumners: Can the `function (match, p1) {}` be defined as a top level utility so that it doesn't have to be recreated every time `prettifyMessage` is invoked?

shape-ideas: > Can you add some escaping support and tests?
> So that somebody could use `\{{level}\} - \{{pid}\}`.

Thanks for your review.
Since `\{' === '{'`, I will update the regex `/{([^}]+)}/` => `/{([^{}]+)}/` for supporting `{{level}}` => `{INFO}`.

I think there is no necessary for escaping, cause the format string must be `\\{{level}\\}`.

shape-ideas: > Can the `function (match, p1) {}` be defined as a top level utility so that it doesn't have to be recreated every time `prettifyMessage` is invoked?

The replacer function needs capture the log, may I ask how to hoist the function? Thanks.

index.js
shape-ideas: Support to a log property can be used by `messageFormat` and `ignore` at the same time.

lib/utils.js
shape-ideas: > Can the `function (match, p1) {}` be defined as a top level utility so that it doesn't have to be recreated every time `prettifyMessage` is invoked?

The replacer function needs capturing the log, may I ask how to hoist the function? Thanks.

lib/utils.js
```diff
@@ -186,6 +186,8 @@ function prettifyLevel ({ log, colorizer = defaultColorizer }) {
  * @param {object} input.log The log object with the message to colorize.
  * @param {string} [input.messageKey='msg'] The property of the `log` that is the
  * message to be prettified.
+ * @param {string} [input.messageFormat=false] An option defines the output of
+ * msg. Default: false
```
jsumners:
```suggestion
```

lib/utils.js
```diff
@@ -186,6 +186,8 @@ function prettifyLevel ({ log, colorizer = defaultColorizer }) {
  * @param {object} input.log The log object with the message to colorize.
  * @param {string} [input.messageKey='msg'] The property of the `log` that is the
  * message to be prettified.
+ * @param {string} [input.messageFormat=false] An option defines the output of
```
jsumners:
```suggestion
 * @param {string} [input.messageFormat=undefined] A format string that defines how the
 *  logged message should be formatted, e.g. `'{level} - {pid}'`.
```

lib/utils.js
```diff
@@ -194,7 +196,16 @@ function prettifyLevel ({ log, colorizer = defaultColorizer }) {
  * key is not a string, then `undefined` will be returned. Otherwise, a string
  * that is the prettified message.
  */
-function prettifyMessage ({ log, messageKey = MESSAGE_KEY, colorizer = defaultColorizer }) {
+function prettifyMessage ({ log, messageFormat, messageKey = MESSAGE_KEY, colorizer = defaultColorizer }) {
+  if (messageFormat) {
+    const message = String(messageFormat).replace(/{([^{}]+)}/g, function (match, p1) {
```
jsumners: Maybe `.replace(/{([^{}]+)}/g, (match, p1) => p1 && log[p1] ? log[p1] : '')`.

lib/utils.js
jsumners: I'm fine with the function as-is. Moving it out of scope would require a `.bind`. And switching to an arrow function has no tangible benefit -- https://jsperf.com/anon-vs-arrow/1

PR #89, opened 2019-11-26
Prettify `caller` property

miguelmota:
If the `caller` property exists (like from [pino-caller](https://github.com/pinojs/pino-caller)), include it in the info line. 

Screenshot from 2019-11-25 17-57-38

### DIFF
--- a/test/basic.test.js
+++ b/test/basic.test.js
@@ -598,5 +598,34 @@ test('basic prettifier tests', (t) => {
     t.is(arst, `[${epoch}] INFO  (on ${hostname}): hello world\n`)
   })
 
+  t.test('prettifies trace caller', (t) => {
+    t.plan(1)
+    const traceCaller = (instance) => {
+      const { symbols: { asJsonSym } } = pino
+      const get = (target, name) => name === asJsonSym ? asJson : target[name]
+
+      function asJson (...args) {
+        args[0] = args[0] || {}
+        args[0].caller = '/tmp/script.js'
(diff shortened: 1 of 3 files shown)


mcollina:
Thanks! can you add a unit test?

miguelmota:
@mcollina unit test added. Let me know what you think!

mcollina:
It seems CI is failing, can you check?

miguelmota:
@mcollina whoops forgot to run linter; all good now

test/basic.test.js
```diff
@@ -598,5 +598,34 @@ test('basic prettifier tests', (t) => {
...
+      return new Proxy(instance, { get })
+    }
+
+    const pretty = prettyFactory()
+    const log = traceCaller(pino({}, new Writable({
+      write (chunk, enc, cb) {
+        const formatted = pretty(chunk.toString())
+        t.is(
+          formatted,
+          `[${epoch}] INFO  (${pid} on ${hostname})</tmp/script.js>: foo\n`
```
jsumners: Why wouldn't there be a ` ` between `)<`?

miguelmota: It's more compact but have no strong preference. Can add a space inbetween

jsumners: A prettifier isn't about compactness, it's about legibility. A space would be more appropriate.

miguelmota: Will do 👍

Issue #91, opened 2019-12-09
Unable to customize "level" key


wraithy:
Pino allows changing the level key via the `changeLevelName` option.

Pino-Pretty already supports `--messageKey`, so it should probably also support `--changeLevelName`.

mcollina:
Would you like to send a PR to add that option?

wraithy:
@mcollina Yes, working on that now.

PR #92, opened 2019-12-09
Add --changeLevelName option

wraithy:
This would resolve https://github.com/pinojs/pino-pretty/issues/91

I've named the option `changeLevelName` to be in line with the option in pino.
I wasn't too sure about the short form for `--changeLevelName`, since both `-c` and `-l` were already used, so right now there is no short form.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -62,6 +62,8 @@ pino app.js | pino-pretty
   error like objects. Default: `err,error`.
 - `--messageKey` (`-m`): Define the key that contains the main log message.
   Default: `msg`.
+- `--levelKey` (`--levelKey`): Define the key that contains the level of the log.
+  Default: `level`.
 - `--messageFormat` (`-o`): Format output of message, e.g. `{level} - {pid}` will output message: `INFO - 1123`
   Default: `false`
 - `--timestampKey` (`-m`): Define the key that contains the log timestamp.
@@ -129,6 +131,7 @@ with keys corresponding to the options described in [CLI Arguments](#cliargs):
   errorProps: '', // --errorProps
   levelFirst: false, // --levelFirst
   messageKey: 'msg', // --messageKey
+  levelKey: 'level', // --levelKey
   messageFormat: false // --messageFormat
   timestampKey: 'time', // --timestampKey
   translateTime: false, // --translateTime
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -7,6 +7,7 @@ const {
   DATE_FORMAT,
   ERROR_LIKE_KEYS,
   MESSAGE_KEY,
+  LEVEL_KEY,
   TIMESTAMP_KEY,
   LOGGER_KEYS
 } = require('./constants')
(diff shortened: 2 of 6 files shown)


wraithy:
@mcollina 
I also didn't update the version in package.json.

Would this be a minor or patch version?

mcollina:
Don’t worry about it, I’ll update when I release.

davidmarkclements:
as a general rule, never worry about the package version - that should always be on the maintainer imo

Readme.md
```diff
@@ -62,6 +62,8 @@ pino app.js | pino-pretty
   error like objects. Default: `err,error`.
 - `--messageKey` (`-m`): Define the key that contains the main log message.
   Default: `msg`.
+- `--changeLevelName` (`--changeLevelName`): Define the key that contains the level of the log.
```
davidmarkclements: in keeping with the other flags let's just make this `--levelKey`

Readme.md
```diff
@@ -129,6 +131,7 @@ with keys corresponding to the options described in [CLI Arguments](#cliargs):
   errorProps: '', // --errorProps
   levelFirst: false, // --levelFirst
   messageKey: 'msg', // --messageKey
+  changeLevelName: 'level', // --changeLevelName
```
davidmarkclements: `levelKey`

wraithy:
I named it `changeLevelName` for parity with [the option in pino](https://github.com/pinojs/pino/blob/master/docs/api.md#changelevelname-string).

I must admit though, I much prefer `levelKey`.

davidmarkclements:
ohhhh shit. I didn't realise that had slipped into pino core. So sorry in that case I think the flag has to match :/ 

@mcollina : we should rename this in v6

wraithy:
@davidmarkclements Cool, I've reverted back to `--changeLevelName`.

jsumners:
> we should rename this in v6

Add an issue?

lib/utils.js
```diff
@@ -166,17 +167,19 @@ function prettifyErrorLog ({
  * string for that level if so.
  *
  * @param {object} input
- * @param {object} input.log The log object which should have a `level` property.
+ * @param {object} input.log The log object.
  * @param {function} [input.colorizer] A colorizer function that accepts a level
  * value and returns a colorized string. Default: a no-op colorizer.
+ * @param {string} levelKey The key to find the level under. Will default to
```
jsumners:
```suggestion
 * @param {string} [levelKey='level'] The key to find the level under.
```

lib/utils.js
```diff
@@ -166,17 +167,19 @@ function prettifyErrorLog ({
...
+ * 'level'.
```
jsumners:
```suggestion
```

lib/utils.js
wraithy: I've made this change 👍

wraithy:
Wow, I've just realised this was never actually merged. @jsumners  I've tweaked that @param comment like you suggested.

Could we get this merged?

Readme.md
```diff
@@ -62,6 +62,8 @@ pino app.js | pino-pretty
   error like objects. Default: `err,error`.
 - `--messageKey` (`-m`): Define the key that contains the main log message.
   Default: `msg`.
+- `--changeLevelName` (`--changeLevelName`): Define the key that contains the level of the log.
+  Default: `level`.
```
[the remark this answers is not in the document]
davidmarkclements: @mcollina I suggested this be changed to levelKey, the reason it's called `changeLevelName` is because we have that option in the programmatic API.  https://github.com/pinojs/pino/blob/master/docs/api.md#changelevelname-string

I think that's a terrible name in the API and we should deprecate it, alias to `levelKey`, and then remove the alias in v6

If we do this, then this PR should update to `levelKey`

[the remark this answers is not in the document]
davidmarkclements: see https://github.com/pinojs/pino-pretty/pull/92#discussion_r356183447

mcollina:
@wraithy can you please follow the suggestions from @davidmarkclements? Thanks

wraithy:
> @wraithy can you please follow the suggestions from @davidmarkclements? Thanks

Sure. From what I understand, the pattern of deprecation and alias etc before removal will apply only to the main repo and not `pino-pretty`. All I've done now is directly rename the option on this PR.

PR #95, opened 2020-01-05
Allow nested key in messageFormat

kelvien:
The existing `messageFormat` does not allow us to print a nested key from an object. i.e:
```
{
  level: 20,
  pid: 123,
  req: {
    url: "localhost:3000/favicon.ico"
  }
}
```
With `--messageFormat '{level} - {pid} - {req.url}'`, pino-pretty will print `20 - 123 - `
With the proposed change, pino-pretty will print `20 - 123 - localhost:3000/favicon.ico`


kelvien:
Added a unit test - @mcollina

Zirak:
This is a good feature and this is just my 2c, however the implementation may be overly simple. For instance:

```js
pino.info({
    'a.b': 4,
    a: {
        b: 7,
    },
});
```

With this PR, formatting `{a.b}` will print out `7`, which is fine on its own, but there's no way of accessing `4`.
To allow accessing both, the more difficult-to-parse subscript notation is needed, something like `{[a.b]}` or `{'a.b'}` or similar, mirroring the difference between `obj.a.b` and `obj['a.b']`.

What do you think?

davidmarkclements:
I think this feature could quickly escalate into something very complex - because we're starting to create a template DSL. I think in retrospect `messageFormat` should have been a function in the first place, but would definitely accept a PR now to make it a function now. We'll continue to support just top level keys for simple `messageFormat` and then you can use a function to access nested properties and do any other type of transform you want

davidmarkclements:
sorry sorry - I forgot this was the prettifier. Ok then if we're going to do it we should do it properly. Use mustache or something.

jsumners:
Or use https://www.npmjs.com/package/lodash.get to interpret the dsl.

kaareal:
Wondering if wouldn't be better to just the native templates syntax https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals.
As no learning should be needed for that. 

Naturally one would need to find a library or implement own that matches the native implementation, but shouldn't be too hard.

Also wondering if it's possible to make the `messageFormat` work with something like `customPrettifiers` which could avoid complex templating => move the logic out of the template and into the `customPrettifiers`.
 
I am saying something like the customPrettifiers because the customPrettifiers doesn't work with nested objects.

hao5743:
image
I faced the same problem. Is this problem had solutions now ? Can somebody help me ?

jsumners:
@hao5743: I think the issue discussion is available for anyone to read. What is unclear to you?

dublx:
Hi all

After looking at using template strings as suggested by @kaareal it was found there is an issue if a property name used in `messageFormat` is not a key in `log`, as it will fail the operation; Parsing `messageFormat` for all property names and cross-checking to `log` keys is overly complex and error prone because it would be possible that messageFormat is not just a simple property name but could instead be any valid expression allowed in a template literals.

For anyone curious, support of tagged template literals was tested using this code:
```js
function prettifyMessage ({ log, messageFormat, messageKey = MESSAGE_KEY, colorizer = defaultColorizer }) {
  if (messageFormat) {
    const formatMessage = function () {
      const func = new Function(...Object.keys(log), 'return `' + messageFormat + '`;') // eslint-disable-line no-new-func
      return func(...Object.values(log))
    }
    try {
      const message = formatMessage()
      return colorizer.message(message)
    } catch (e) {
      return undefined
    }
  }
  if (messageKey in log === false) return undefined
  if (typeof log[messageKey] !== 'string') return undefined
  return colorizer.message(log[messageKey])
}
```

Alternatively using a third-party library as suggested above would probably be a poor option regarding additional support and maintenance.

As such it would be recommended we land this PR despite the drawback identified by @Zirak in https://github.com/pinojs/pino-pretty/pull/95#issuecomment-571502494

Please let us know your feedback.

mcollina:
cc @davidmarkclements can we land this as is? I think this would be an improvement anyway.

lib/utils.js
```diff
@@ -201,10 +201,12 @@ function prettifyLevel ({ log, colorizer = defaultColorizer, levelKey = LEVEL_KE
 function prettifyMessage ({ log, messageFormat, messageKey = MESSAGE_KEY, colorizer = defaultColorizer }) {
   if (messageFormat) {
     const message = String(messageFormat).replace(/{([^{}]+)}/g, function (match, p1) {
-      if (p1 && log[p1]) {
-        return log[p1]
-      }
-      return ''
+      return p1.split('.').reduce(function (prev, curr) {
```
jsumners, line 204:
```suggestion
      // Parse nested key access, e.g. `{keyA.subKeyB}`.
      return p1.split('.').reduce(function (prev, curr) {
```

dublx:
> cc @davidmarkclements can we land this as is? I think this would be an improvement anyway.

Ok to merge?

Zirak:
@dublx You're right that your implementation of supporting template literals wholesale raises a few concerns.

However, you don't have to support _all_ of template literals. You can support a subset with just getters. That would have problems of its own, since there's the user expectation of what should happen with a DSL they know, and pino-pretty's implementation of that DSL.

That's why I think the stricter subset of `foo.bar` and `foo['bar']` is more palpable. The lodash implementation is ~50 LoC, and the only thing lost is the initial `$` (`${foo.bar}` vs. `{foo.bar}`).

However I'll also side with you and @mcollina against my past self. Merging this in doesn't close the opportunity of implementing the "proper" property matching.

davidmarkclements (approved):
yeah sounds good to me :+1:

davidmarkclements:
once conflicts are we resolved we're good to merge

lib/utils.js
```diff
@@ -203,13 +203,18 @@ function prettifyLevel ({ log, colorizer = defaultColorizer, levelKey = LEVEL_KE
...
     const message = String(messageFormat).replace(/{([^{}]+)}/g, function (match, p1) {
+      // return log level as string instead of int
       if (p1 === levelLabel && log[LEVEL_KEY]) {
         return LEVELS[log[LEVEL_KEY]]
       }
-      if (p1 && log[p1]) {
-        return log[p1]
-      }
-      return ''
+
```
mcollina, line 210:
```suggestion

```

PR #103, opened 2020-03-31
100% code coverage

mcollina:
This also update all dependencies  and fixes a few error discovered while bringing the code coverage to 100%.

### DIFF
--- a/index.js
+++ b/index.js
@@ -69,11 +69,6 @@ module.exports = function prettyFactory (options) {
       log = inputData
     }
 
-    // Short-circuit for spec allowed primitive values.
-    if ([null, true, false].includes(log) || Number.isFinite(log)) {
-      return `${log}\n`
-    }
-
     if (search && !jmespath.search(log, search)) {
       return
     }
--- a/test/cli-rc.test.js
+++ b/test/cli-rc.test.js
@@ -22,7 +22,7 @@ test('cli', (t) => {
     const configFile = path.join(tmpDir, 'pino-pretty.config.js')
     fs.writeFileSync(configFile, 'module.exports = { translateTime: true }')
     const env = { TERM: 'dumb' }
-    const child = spawn(process.argv0, [bin], { env, cwd: tmpDir })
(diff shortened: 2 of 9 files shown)


index.js
```diff
@@ -69,11 +69,6 @@ module.exports = function prettyFactory (options) {
       log = inputData
     }
 
-    // Short-circuit for spec allowed primitive values.
-    if ([null, true, false].includes(log) || Number.isFinite(log)) {
-      return `${log}\n`
-    }
-
```
jsumners, line 76: Why remove this? It's checking for the input being literally `null`, `true`, `false`, or a number and simply passing it along to the output. It's covered by these tests https://github.com/pinojs/pino-pretty/blob/ec3c5854a4b42f9fd9005bfa4a94756e484f8732/test/basic.test.js#L354-L386

test/cli-rc.test.js
```diff
@@ -22,7 +22,7 @@ test('cli', (t) => {
     const configFile = path.join(tmpDir, 'pino-pretty.config.js')
     fs.writeFileSync(configFile, 'module.exports = { translateTime: true }')
     const env = { TERM: 'dumb' }
-    const child = spawn(process.argv0, [bin], { env, cwd: tmpDir })
+    const child = spawn(process.argv[0], [bin], { env, cwd: tmpDir })
```
jsumners, line 25: Why these changes? `argv0` has been present since 6 -- https://nodejs.org/dist/latest-v12.x/docs/api/process.html#process_process_argv0

index.js
mcollina: Because it's never triggered. It's currently handled in https://github.com/pinojs/pino-pretty/blob/ec3c5854a4b42f9fd9005bfa4a94756e484f8732/index.js#L65

test/cli-rc.test.js
mcollina: Because `node test/cli-rc.test.js` was failing on my machine (Mac) on Node 12 if run without `tap`. I'm not sure why and I did not investigate.

Issue #105, opened 2020-04-13
Journalctl (not an issue)


Elexy:
I just wanted to share how to get pretty logs from node running under systemd:
`sudo journalctl -u monitor -f -o cat | pino-pretty`

from the journalctl manual:
> cat 
generates a very terse output, only showing the actual message of each journal entry with no metadata, not even a timestamp.

mcollina:
Nice @elexy! I think we should get this into https://github.com/pinojs/pino/blob/master/docs/help.md, if wouldn't mind to do a PR.

PR #106, opened 2020-04-24
Systemd example

Elexy:
PR in refence to https://github.com/pinojs/pino-pretty/issues/105

### DIFF
--- a/docs/help.md
+++ b/docs/help.md
@@ -1,0 +1,24 @@
+<a id="Systemd"></a>
+## Systemd example
+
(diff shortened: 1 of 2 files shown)


docs/help.md
```diff
@@ -0,0 +1,24 @@
+<a id="Systemd"></a>
+## Systemd example
+
+If you run you Node.js process in [Systemd](https://www.freedesktop.org/wiki/Software/systemd/) and you examine your logs with [journalctl](https://www.freedesktop.org/software/systemd/man/journalctl.html) some data will be duplicated. You can use a combination of `Journalctl` options and `pino-pretty` options to shape the output.
```
jsumners, line 4:
```suggestion
If you run your Node.js process via [Systemd](https://www.freedesktop.org/wiki/Software/systemd/) and you examine your logs with [journalctl](https://www.freedesktop.org/software/systemd/man/journalctl.html) some data will be duplicated. You can use a combination of `journalctl` options and `pino-pretty` options to shape the output.
```

Issue #116, opened 2020-07-10
Ignoring the timestampKey


BenGale:
I want to use the option to set `timestampKey` to a different field called `@timestamp`. This works as expected, but I also want to add it to the ignore list because I don't want each log message to have the timestamp outputted underneath as well.

The problem is when I add it to the ignore list it no longer gets picked up as the timestampKey. Is this the way its expected to work?

My current options seem to be:
![image](https://user-images.githubusercontent.com/1517604/87147055-d8bdda00-c2a3-11ea-95f7-3a850a79c4c5.png)
or
![image](https://user-images.githubusercontent.com/1517604/87147078-e7a48c80-c2a3-11ea-9add-a2400a2dc928.png)

PR #117, opened 2020-07-10
Prettify Time Before Filtering

BenGale:
This allows the prettifiedTime function run before any of the properties are filtered from the log object. If you are using a timestampKey that is also ignored it will still work.

Issue described in: https://github.com/pinojs/pino-pretty/issues/116

### DIFF
--- a/test/cli.test.js
+++ b/test/cli.test.js
@@ -101,7 +101,7 @@ test('cli', (t) => {
   })
 
   t.test('uses specified timestampKey', (t) => {
-    t.plan(1);
+    t.plan(1)
     const env = { TERM: 'dumb' }
     const child = spawn(process.argv[0], [bin, '--timestampKey', '@timestamp'], { env })
     child.on('error', t.threw)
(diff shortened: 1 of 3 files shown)


BenGale:
Good point. Do the tests I've added there look ok to you?

test/cli.test.js
```diff
@@ -100,5 +100,31 @@ test('cli', (t) => {
...
+    const env = { TERM: 'dumb' }
+    const child = spawn(process.argv[0], [bin, '--timestampKey', '@timestamp', '--ignore', '@timestamp'], { env })
+    child.on('error', t.threw)
+    child.stdout.on('data', (data) => {
+      t.is(data.toString(), '[1522431328992] INFO : hello world\n')
+    })
+    const logLine = '{"level":30,"@timestamp":1522431328992,"msg":"hello world"}\n'
+    child.stdin.write(logLine)
+    t.tearDown(() => child.kill())
+  });
```
jsumners: Thank you. These look good for covering what the CLI does. I'd also like to see the behavior covered in the factory tests, e.g.:

https://github.com/pinojs/pino-pretty/blob/62a5f01d31aef4ca8ff69a08180080bd88506c2c/test/basic.test.js#L603-L615

BenGale: No problem, I've added those too.

PR #120, opened 2020-07-14
Skip timestampKey in Prettified Object

BenGale:
@mcollina This is related to: https://github.com/pinojs/pino-pretty/pull/119

I took another look at a better way for this to be handled and it seems like there is already something handling overrides for the messageKey and the levelKey, so I have extended that to also support timestampKey. Apologies for not spotting this on the first fix PR.

There are tests to cover this and make sure it isn't outputted in the prettified object.

### DIFF
--- a/index.js
+++ b/index.js
@@ -133,7 +133,7 @@ module.exports = function prettyFactory (options) {
       })
       line += prettifiedErrorLog
     } else {
-      const skipKeys = [messageKey, levelKey].filter(key => typeof log[key] === 'string')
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -133,7 +133,7 @@ module.exports = function prettyFactory (options) {
       })
       line += prettifiedErrorLog
     } else {
-      const skipKeys = [messageKey, levelKey].filter(key => typeof log[key] === 'string')
+      const skipKeys = [messageKey, levelKey, timestampKey].filter(key => typeof log[key] === 'string' || typeof log[key] === 'number')
```
jsumners, line 136: Why the extra type check?

BenGale: The time comes through as a number.

PR #123, opened 2020-08-17
suport nested key value for messageFormat

hao5743:
Fix to suport nested value. Only when not specific key matched.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -199,15 +199,19 @@ function prettifyLevel ({ log, colorizer = defaultColorizer }) {
 function prettifyMessage ({ log, messageFormat, messageKey = MESSAGE_KEY, colorizer = defaultColorizer }) {
   if (messageFormat) {
     const message = String(messageFormat).replace(/{([^{}]+)}/g, function (match, p1) {
-      if (p1 && log[p1]) {
+      if (!p1) return ''
+      if (log[p1] && typeof log[p1] !== 'object') {
         return log[p1]
       }
-      return p1.split('.').reduce(function (prev, curr) {
-        if (prev && prev[curr]) {
-          return prev[curr]
+      let obj = log
+      for (const k of p1.split('.')) {
+        if (!obj[k]) return ''
+        if (typeof obj[k] !== 'object') {
+          return obj[k]
         }
-        return ''
-      }, log)
+        obj = obj[k]
+      }
+      return ''
     })
     return colorizer.message(message)
   }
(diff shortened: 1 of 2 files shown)


mcollina:
Please lint the code, CI is failing.

lib/utils.js
```diff
@@ -202,7 +202,12 @@ function prettifyMessage ({ log, messageFormat, messageKey = MESSAGE_KEY, colori
       if (p1 && log[p1]) {
         return log[p1]
       }
-      return ''
+      return p1.split('.').reduce(function (prev, curr) {
+        if (prev && prev[curr]) {
```
mcollina, line 206: this will have some problems if we have `curr` as a `length` property or anything else that is present on a string. I think we should use a for loop instead of a reduce and just break/return early

test/lib/utils.public.test.js
```diff
@@ -101,6 +101,16 @@ tap.test('prettifyMessage', t => {
...
 
+  t.test('returns nested value in object', async t => {
+    const str = prettifyMessage({ log: { msg: 'foo', req: { id: '123' } }, messageFormat: '{req.id}' })
+    t.is(str, '123')
+  })
+
+  t.test('returns nested value in object only if not specific key', async t => {
+    const str = prettifyMessage({ log: { msg: 'foo', 'req.id': '567', req: { id: '123' } }, messageFormat: '{req.id}' })
+    t.is(str, '567')
+  })
```
mcollina, line 112: can you please add a test where `{object.with.deep.property}` is not present completely?

hao5743: '[object Object]' will get.  Do you have any suggestions for this?

mcollina: I think it should be empty instead

lib/utils.js
hao5743: fixed it to loop and return early

test/lib/utils.public.test.js
hao5743: return '' instead

lib/utils.js
```diff
@@ -199,9 +199,17 @@ function prettifyLevel ({ log, colorizer = defaultColorizer }) {
 function prettifyMessage ({ log, messageFormat, messageKey = MESSAGE_KEY, colorizer = defaultColorizer }) {
   if (messageFormat) {
     const message = String(messageFormat).replace(/{([^{}]+)}/g, function (match, p1) {
-      if (p1 && log[p1]) {
+      if (p1 && log[p1] && typeof log[p1] !== 'object') {
```
jsumners, line 202: Please use `isObject` https://github.com/pinojs/pino-pretty/blob/852fdf707c1a57627a3f523b0cec8824e5d24ab4/lib/utils.js#L78-L80

hao5743: Had fixed it. The issue is #124

davidmarkclements (approved):
fair enough :+1:

davidmarkclements (changes requested):
sorry wrong PR - see @jsumners request for change

PR #130, opened 2020-09-06
Avoid adding spaces at the beginning of the line, fixes #129

### LINKED ISSUE #129: Leading space for ignore time and level
If I use `--ignore` tp exclude the time and level, I get an unwanted leading space on each line. For example:

```
echo '{"level":30,"time":1599275372412,"pid":26631,"hostname":"pb","name":"server","msg":"Hello, world!"}' \
  | npx pino-pretty --ignore "time,level"
```

This yields:

```
 (server/26631 on jitterbug): Hello, world!
```

Note the space at the beginning of the line.

I seem to get this space regardless of whether or not I ignore additional keys or which ones. What I'm actually trying to do is display just the message:

```
echo '{"level":30,"time":1599275372412,"pid":26631,"hostname":"pb","name":"server","msg":"Hello, world!"}' \
  | npx pino-pretty --ignore "time,level,name,pid,hostname"
```

Again, this includes a leading space:

```
 Hello, world!
```

This seems like a bug, but I'm not sure. Is there a reason for including the space? I'd much prefer not to have it there.

davesteinberg:
Here is my fix for issue #129. There were two spots in pretty() where it was adding a space, even if the line to that point was empty, so I added tests for and fixed both.

### DIFF
--- a/index.js
+++ b/index.js
@@ -108,7 +108,11 @@ module.exports = function prettyFactory (options) {
     }
 
     if (prettifiedMetadata) {
-      line = `${line} ${prettifiedMetadata}:`
+      if (line.length > 0) {
+        line = `${line} ${prettifiedMetadata}:`
+      } else {
+        line = prettifiedMetadata
+      }
     }
 
     if (line.endsWith(':') === false && line !== '') {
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

PR #137, opened 2020-10-10
Unblock logs on jmespath search error

kelvien:
Right now pino-pretty will break when it encounters any `jmespath.search` error.
Instead of breaking the entire tool while continuously tailing logs, it should continue through processing the rest of the logs and notify users when it can't process the log.

### DIFF
--- a/index.js
+++ b/index.js
@@ -75,7 +75,7 @@ module.exports = function prettyFactory (options) {
         return
       }
     } catch (error) {
-      console.error(`Unable to search current log ${JSON.stringify(log)} due to ${error}`);
+      console.error(`Unable to search current log ${JSON.stringify(log)} due to ${error}`)
       return
     }
 
--- a/index.js
+++ b/index.js
@@ -70,7 +70,11 @@ module.exports = function prettyFactory (options) {
       log = inputData
     }
 
-    if (search && !jmespath.search(log, search)) {
+    try {
+      if (search && !jmespath.search(log, search)) {
+        return
+      }
+    } catch (error) {
       return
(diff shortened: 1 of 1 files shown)


mcollina:
It seems code coverage is lowering with this, could you cover the missing lines?

jimmiehansson:
I'm actually more concerned about the exploitation of this as there's nothing in jmespath preventing an excursion of a range error or amplification unless properly preventing the callstack from growing and processing gets out of hand.

Or do I misunderstand this completely?

E.g.

```javascript
    const buff8 = new ArrayBuffer(8)
    const noop = () => noop(buff8);
    const pretty = prettyFactory({ search: noop() })
```

jsumners:
This is developer targeted module. We do not recommend, nor support, its usage in any sort of "production" environment. Any assertion of an "exploitation" is therefore spurious in my opinion.

jimmiehansson:
> This is developer targeted module. We do not recommend, nor support, its usage in any sort of "production" environment. Any assertion of an "exploitation" is therefore spurious in my opinion.

Fair enough! Its not of concern you may dismiss my comment. I'll see if I can get some time for the coverage for this one

jsumners:
@jimmiehansson please fork from this branch if you do so. We should take care to include the git history for proper attribtion.

jimmiehansson:
> @jimmiehansson please fork from this branch if you do so. We should take care to include the git history for proper attribution.

Oh, sure. Force of habit, I'll fix that

mcollina:
Close with no activity.

PR #139, opened 2020-12-09
fix: wrong array being iterated

rohit-gohri:
Saw this error while seeing the code as reference for https://github.com/pinojs/pino-http-print/issues/9

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -308,7 +308,7 @@ function prettifyObject ({
   let result = ''
 
   const keysToIterate = objectKeys.filter(k => keysToIgnore.includes(k) === false)
-  for (var i = 0; i < objectKeys.length; i += 1) {
+  for (var i = 0; i < keysToIterate.length; i += 1) {
     const keyName = keysToIterate[i]
     const keyValue = input[keyName]
 


jsumners (commented):
Can we cover this with a unit test?

rohit-gohri:
> Can we cover this with a unit test?

I'm not sure how since this wasn't causing any errors/bugs, because of this check just below: https://github.com/rohit-gohri/pino-pretty/blob/585c0bb7999b6cfc52e06b1e951e267f8758b809/lib/utils.js#L315

jsumners:
Compare an object that doesn't iterate correctly prior to the change with a known good output.

rohit-gohri:
> Compare an object that doesn't iterate correctly prior to the change with a known good output.

That's what I'm saying, even though the wrong array was being iterated the output was already correct because the value was compared against undefined before doing anything with it:

https://github.com/rohit-gohri/pino-pretty/blob/585c0bb7999b6cfc52e06b1e951e267f8758b809/lib/utils.js#L310-L316

mcollina:
If it access the wrong index, it could result printing the wrong value maybe? The test could verify if there are no spurious keys.

mcollina:
Close with no activity.

Issue #140, opened 2020-12-15
Lots of spaces after TRACE, but not so after INFO


emonddr:
I am using fastify with pino and pino pretty.

I see lots of spaces after the TRACE keyword, but not after the INFO keyword.

```
[2020-12-15T20:33:16:892Z] INFO : Server listening at http://127.0.0.1:3000
[2020-12-15T20:33:28:275Z] TRACE        : [c0c93293-5093-40bf-8d32-76a61287fba7] \/ Entering [POST /someEndPoint]
```

I tried creating a custom prettifier for the 'level' field, but it is never called.

Any insight is appreciated.

thx

:)

jsumners:
See https://github.com/pinojs/pino-pretty/pull/135

emonddr:
@jsumners , that for the link.

 I am still seeing the large gap between the TRACE string and the beginning of the string I actually sent to fastify.log.trace() .  what version of pino-pretty contains the fix above?

I am using
```
 "pino-pretty": "^4.3.0",
```

I also have these options set

```
level: 'trace',
  prettyPrint: {
    ignore: 'pid,hostname',
    translateTime: `UTC:yyyy-mm-dd'T'HH:MM:ss:l'Z'`,
  },
```

emonddr:
Also, I don't have a custom prettifier for 'level' defined when this large gap is occurring.

jsumners:
I am saying that PR #135 introduced tabs as alignment tokens. So this is likely a bug. Personally, I'd rather just revert back to not trying to align these fields. But if you, @Zirak, and @dublx would like to collaborate on a fix that retains the formatting, that would be okay as well.

arthurw1:
I'm observing something similar too. See below:

![image](https://user-images.githubusercontent.com/75303606/103052127-561ba380-45d3-11eb-8ec6-1634a2ded482.png)

mcollina:
This issue needs somebody to work on it, volunteers are welcome :).

arthurw1:
@mcollina I can take this on!

mcollina:
go for it!

PR #135, opened 2020-10-01
Use tab to align output of all levels

dublx:
Improves on https://github.com/pinojs/pino-pretty/pull/47

Same visual result but the trailing spaces in INFO and WARN constants were removed.
```sh
$ node example.js | pino-pretty
[1601561263881] INFO	 (39775 on nf.lan): Ini
[1601561263881] ERROR	 (39775 on nf.lan): Miny
[1601561263881] WARN	 (39775 on nf.lan): Mo!
```

PR #141, opened 2020-12-30
Fix/remove extra space

arthurw1:
This PR addresses #140 

## Update
This PR removes all alignment spaces for log level constants.

---

The change here is simply to revert #135 

The extra spaces highlighted in #140 was caused by using `\t` as delimiter. 

Result:
image


mcollina (approved):
lgtm

jsumners:
This is basically creating churn between space formatting and tab formatting. @Zirak introduced space formatting because they found it easier to read the formatted output. @dublx converted it to tab based formatting for more consistency. I think we need a better solution than alternating between these two every few months.

mcollina:
@jsumners wdyt? should we land this or not?

jsumners:
> @jsumners wdyt? should we land this or not?

I really don't think so. Not until the parties involved that have been interested in this formatting speak up (@Zirak and @dublx).

My opinion is we completely remove this alignment formatting feature completely if no one is able to propose a solution that works as everyone expects. It seems to me that the issue is down to the terminal emulator and font being used and how the present `\t` and ` `.

I just don't know how we can satisfy all desires here easily. Reverting the `\t` change will cause #135 to be an issue again.

dublx:
Aye, I totally agree. 
I'm burned out of time to work on this unfortunately but in terms of choice i'd go with investing time to make \t work well for most people.

mcollina:
I think we should revert the feature.

jsumners:
@arthurw1 would you like to create a PR that completely reverts the space formatting feature?

arthurw1:
@jsumners Yes I can take that on. I will simply make the changes on top of this current branch if that's ok.

jsumners:
That is fine.

arthurw1:
@jsumners @mcollina PR updated based on latest decision

mcollina (approved):
lgtm

jsumners (approved):
LGTM

jsumners:
@mcollina do we think this is patch, minor, or major? I think I could make an argument for any one of them.

mcollina:
I would land this as a minor. If you are parsing this manually you are doing it wrong.

PR #144, opened 2021-01-24
Complement PR #139 with tests

jimmiehansson:
Description
This PR complements existing PR #139 with tests, implementing the same working solution.
Fork corrected to initial branch history

### DIFF
--- a/test/lib/utils.public.test.js
+++ b/test/lib/utils.public.test.js
@@ -233,6 +233,26 @@ tap.test('prettifyObject', t => {
     t.is(str, '    foo: "bar"\n')
   })
 
+  t.test('iterates input and return result from exclusion by ignored keys undefined', async t => {
+    const keysToIgnore = ['undefined']
+    const fixture = {
+      undefined: undefined,
+      foo: { bar: 'baz' }
+    }
+    const res = prettifyObject({ input: fixture, skipKeys: keysToIgnore })
+    t.is(res, '    foo: {\n      "bar": "baz"\n    }\n')
+  })
+
+  t.test('iterates input and return result from exclusion by ignored keys foo', async t => {
+    const keysToIgnore = ['foo']
+    const fixture = {
+      undefined: undefined,
+      foo: { bar: 'baz' }
+    }
+    const res = prettifyObject({ input: fixture, skipKeys: keysToIgnore })
+    t.is(res, '')
+  })
+
   t.test('works with error props', async t => {
     const err = Error('Something went wrong')
     const serializedError = {


mcollina:
I don't think this could compliment #139, as the tests are passing in master without that fix.

jsumners:
Other than Matteo's comment, if we were to merge this PR I do not believe the #139 changes would get merged in. I would expect to see the original PR's commit(s) here along with your own. The workflow would look something like:

```sh
$ git clone <your fork of pino.git>
$ git remote add pino-rohit https://github.com/rohit-gohri/pino-pretty.git
$ git checkout -b my-feature
$ git pull pino-rohit patch-1
$ # make changes
$ git add . && git commit -m 'Add my tests'
```

jimmiehansson:
Alright, that makes more sense. Thanks for clarification @jsumners I'll have it updated.

> I don't think this could compliment #139, as the tests are passing in master without that fix.

@mcollina How would you like for me to proceed? Being that it passes even without the fix, doesn't consent with the problem existing to that case

mcollina:
In order to trigger the mentioned case, you need to add more keys to the object. Essentially you need to cause a situation where `keysToIterate[i]` and `objectKeys[i]` have different values. https://github.com/pinojs/pino-pretty/pull/139/files#diff-5dfe38baf287dcf756a17c2dd63483781b53bf4b669e10efdd01e74bcd8e780aR312

jimmiehansson:
Oh I see, that makes more sense. I'll have more time for it tomorrow evening and see if I can get an update in place to actually reproduce it

jimmiehansson:
Hey guys, not ignoring the task but got a terrible neck brace / headache at the end of last week so riding this out first

jsumners:
Take your time.

jimmiehansson:
So this one has been stalled, been in a cyclone of work. I'll look further into it during the evening Today or Tomorrow, if anyone has more time spared to see if its reproducible please go ahead!

jimmiehansson:
Alright, if I've understood this case correctly we would need the lists to behave differently than what is expected of `includes` and the only way according to my knowledge, is to input non-matching elements that includes won't respect `includes` while performing its matching algorithm.

as in:

```javascript
    const keysToIgnore = [0]
    const fixture = {
      0: 0
    }
    const res = prettifyObject({ input: fixture, skipKeys: keysToIgnore })
```

I've managed to reproduce this by inputting a numeric value to the ignore list, which as expected with `includes` will not return the expected ignored result.

But this is a subject of the ignored list not matching the expectation of a filter, rather than the two lists not being consistent.

Can someone guide me here, or am I simply too tired?

mcollina:
> But this is a subject of the ignored list not matching the expectation of a filter, rather than the two lists not being consistent.

I think this is the same bug, yes.

jimmiehansson:
Alright, thanks! I'll see what I can wrap up

jimmiehansson:
I'll go ahead and wrap this up with a reference to #149

PR #145, opened 2021-01-28
feature: Add a flag to hide object from output (--hideObject)

louisvenne:
This PR adds the flag 'hideObject' that allows users to choose disabling the print of objects on logs (i.e `prettifyObject()` call). I tried to modify the least code as possible to meet this goal.

### DIFF
--- a/test/basic.test.js
+++ b/test/basic.test.js
@@ -684,5 +684,18 @@ test('basic prettifier tests', (t) => {
     log.info({ v: 1 })
   })
 
+  t.test('Hide object `{ key: "value" }` from output when flag `hideObject` is set', (t) => {
+    t.plan(1)
+    const pretty = prettyFactory({ hideObject: true })
+    const log = pino({}, new Writable({
+      write (chunk, enc, cb) {
+        const formatted = pretty(chunk.toString())
+        t.is(formatted, `[${epoch}] INFO\t (${pid} on ${hostname}):\n`)
+        cb()
+      }
+    }))
+    log.info({ key: 'value' })
+  })
+
   t.end()
 })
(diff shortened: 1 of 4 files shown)


test/basic.test.js
```diff
@@ -684,5 +684,18 @@ test('basic prettifier tests', (t) => {
...
+    const pretty = prettyFactory({ hideObject: true })
+    const log = pino({}, new Writable({
+      write (chunk, enc, cb) {
+        const formatted = pretty(chunk.toString())
+        t.is(formatted, `[${epoch}] INFO\t (${pid} on ${hostname}):\n`)
+        cb()
+      }
+    }))
+    log.info({ key: 'value' })
+  })
```
mcollina, line 698: Does it make sense to actually log this line? It says _nothing_ as there is no message and the object is ignored.

I think adding this option would make the output more confusing.

mcollina (changes requested):
lgtm

test/basic.test.js
louisvenne: For test code, I tried to do same thing as the test just above (i.e `log.info({ v: 1 })`) to simulate an basic object that will be ignore from output. I just put `{ key: 'value' }`for the example. Feel free to modify it. But maybe I misunderstood something in your test logic :/

For the option, personally I prefer concise log for development. If I want to debug a endpoint I will log the variable manually. I think this option won't disturb other users as default value is `false` and so change nothing to normal use.

jsumners: @louisvenne change it to `log.info({key: 'value'}, 'hello world')`. Then verify that "hello world" is printed _without_ `{key: 'value'}` being serialized.

louisvenne: done

mcollina (approved):
lgtm

PR #146, opened 2021-01-31
feature: allow the user to set the `messageFormat` option as a function

louisvenne:
This PR allows users to choose to define the `messageFormat` option as a template `string` or a `function`. Like the previous PR, I tried to modify the least code as possible to achieve this goal and not change the normal use of the library.


mcollina (approved):
lgtm

jsumners (approved):
LGTM

PR #148, opened 2021-02-10
Support Winston Log Levels

davidsiu:
This introduces an out of the box experience that allows previous Winston users to map `NOTICE` and `SILLY` log levels. One simply needs to conform to setting the level values of:

- `35` for `NOTICE`
- `5` for `SILLY`


davidsiu:
Whoops wrong repo :P

PR #149, opened 2021-02-13
Add sanity check and complement PR #139 with tests (add history)

jimmiehansson:
**Description**
This PR complements existing PR #139 with tests and a solution for filtering primitives and parsing numbers properly.

* Add number-to-string coercion for ignored keys
* Exclude null references for ignored keys
* Add tests

**OBS:** This PR renders #144 obsolete as it does not contain the proper git history and attribution to @rohit-gohri

Pardon the chaotic history of this PR

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -309,7 +309,7 @@ function prettifyObject ({
 
   const keysToIterate = safeObjectKeys()
   var i = 0
-  for (i; i < objectKeys.length; i += 1) {
+  for (i; i < keysToIterate.length; i += 1) {
     const keyName = keysToIterate[i]
     const keyValue = input[keyName]
 
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -307,13 +307,6 @@ function prettifyObject ({
 
   let result = ''
 
-  function safeObjectKeys () {
-    const safePrims = objectKeys.filter(k => k !== 'null')
-    const strArr = keysToIgnore.join(',').split(',')
-    const sanitized = safePrims.filter(k => strArr.includes(k) === false)
-    return sanitized
-  }
-
   const keysToIterate = safeObjectKeys()
   var i = 0
   for (i; i < objectKeys.length; i += 1) {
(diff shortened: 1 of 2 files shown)


lib/utils.js
```diff
@@ -307,8 +307,16 @@ function prettifyObject ({
 
   let result = ''
 
-  const keysToIterate = objectKeys.filter(k => keysToIgnore.includes(k) === false)
-  for (var i = 0; i < objectKeys.length; i += 1) {
+  function safeObjectKeys () {
```
jsumners, line 310: Could we move this declaration below `return result`? It'll be hoisted appropriately and I think the organization is easier to read.

jimmiehansson: Sure, that makes sense. I've updated the method placement.

rohit-gohri:
Thanks @jimmiehansson for taking this up and sorry I didn't end up adding tests to the original PR, was in the middle of changing jobs.

lib/utils.js
```diff
@@ -307,8 +307,16 @@ function prettifyObject ({
...
+    const safePrims = objectKeys.filter(k => k !== 'null')
+    const strArr = keysToIgnore.join(',').split(',')
+    const sanitized = safePrims.filter(k => strArr.includes(k) === false)
+    return sanitized
+  }
+
+  const keysToIterate = safeObjectKeys()
+  var i = 0
+  for (i; i < objectKeys.length; i += 1) {
```
rohit-gohri, line 312: I couldn't help but notice that this code is again iterating `objectKeys` array but using the index on `keysToIterate` variable. Shouldn't this be :

```suggestion
  for (i; i < keysToIterate.length; i += 1) {
```

jimmiehansson: Thanks for catching this Rohit, this was an unintended reverse during branch changing

mcollina (approved):
lgtm

jsumners (approved):
LGTM

Just need to update the branch.

lib/utils.js
```diff
@@ -349,6 +350,18 @@ function prettifyObject ({
...
+   * Should return a sanitized list of ignored keys
+   *
+   * @returns {iterable}
+   */
+  function safeObjectKeys () {
+    const safePrims = objectKeys.filter(k => k !== 'null')
+    const strArr = keysToIgnore.join(',').split(',')
+    const sanitized = safePrims.filter(k => strArr.includes(k) === false)
+    return sanitized
+  }
```
mcollina, line 364: can you move this out of the main function:

```js
  function safeObjectKeys (objectKeys, strArray) {
    const safePrims = objectKeys.filter(k => k !== 'null')
    const sanitized = safePrims.filter(k => strArr.includes(k) === false)
    return sanitized
  }
```

mcollina (changes requested):

mcollina:
There is also a conflict to fix.

mcollina:
Close with no activity.

PR #157, opened 2021-02-19
chore: add parallel coverage generation; linting

Fdawgs:
Sorry @jsumners, looking through the code trying to figure out why i'm suddenly having issues with `pino-pretty` on Windows, so there may be a few more commits as I go!


mcollina (approved):
lgtm

PR #161, opened 2021-03-02
feat: gray out meta data object when using `--singleLine=true`

### LINKED ISSUE #160: Gray out meta information when using --singleLine option when coloring is enabled
I really like the single-line output introduced in https://github.com/pinojs/pino-pretty/issues/97 🙇🏼🙏🏼 

image

I wonder if we could gray out the meta information by default, when coloring is enabled?

image

I think that makes the logs more readable. 

If you'd like that change, I'd appreciate a pointer on how the change would be done. I looked into `lib/utils.js` but I'm not sure how that change would be implemented, other than requiring `chalk` in this file and making this change

```diff
-       result += stringifySafe(plain)
+       result += chalk.gray(stringifySafe(plain)
```

gr2m:
fixes #160 

Before 

image

After

image

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -3,7 +3,6 @@
 const dateformat = require('dateformat')
 const stringifySafe = require('fast-safe-stringify')
 const defaultColorizer = require('./colors')()
-const chalk = require('chalk')
 const {
   DATE_FORMAT,
   ERROR_LIKE_KEYS,
@@ -308,7 +307,8 @@ function prettifyObject ({
   customPrettifiers = {},
   errorLikeKeys = ERROR_LIKE_KEYS,
   excludeLoggerKeys = true,
-  singleLine = false
+  singleLine = false,
+  colorizer = defaultColorizer
 }) {
   const keysToIgnore = [].concat(skipKeys)
 
@@ -335,7 +335,7 @@ function prettifyObject ({
   if (singleLine) {
     // Stringify the entire object as a single JSON line
     if (Object.keys(plain).length > 0) {
-      result += chalk.grey(stringifySafe(plain))
+      result += colorizer.greyMessage(stringifySafe(plain))
     }
     result += eol
   } else {
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -3,6 +3,7 @@
 const dateformat = require('dateformat')
 const stringifySafe = require('fast-safe-stringify')
 const defaultColorizer = require('./colors')()
+const chalk = require('chalk')
 const {
   DATE_FORMAT,
   ERROR_LIKE_KEYS,
@@ -334,7 +335,7 @@ function prettifyObject ({
   if (singleLine) {
     // Stringify the entire object as a single JSON line
     if (Object.keys(plain).length > 0) {
-      result += stringifySafe(plain)
+      result += chalk.grey(stringifySafe(plain))
     }
     result += eol
   } else {
--- a/index.js
+++ b/index.js
@@ -154,7 +154,8 @@ module.exports = function prettyFactory (options) {
         errorLikeKeys: errorLikeObjectKeys,
         eol: EOL,
         ident: IDENT,
-        singleLine
+        singleLine,
+        colorizer
       })
       line += prettifiedObject
     }
--- a/lib/colors.js
+++ b/lib/colors.js
@@ -11,7 +11,8 @@ const plain = {
   30: nocolor,
   20: nocolor,
   10: nocolor,
-  message: nocolor
+  message: nocolor,
+  greyMessage: nocolor
 }
 
 const chalk = require('chalk')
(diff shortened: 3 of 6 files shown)


gr2m:
I am confused, the tests should not have passed with the changes from 6641ce0, I only pushed the tests changes to show that they are failing, before pushing the fix

gr2m:
Is it possible that on CI the colors are disabled, so testing for colors is impossible?

gr2m:
When running this locally I get

image

image

gr2m:
I'm not sure if this only affects my code, but it looks like tests for colors in stdout won't work in CI because colors are disabled by default. If the code changes from this PR were removed in future, the tests would still pass in CI (but fail locally)

mcollina:
You should use https://github.com/pinojs/pino-pretty/blob/master/lib/colors.js and not chalk directly. You'll be able to test with that.

gr2m:
I'm sorry but I don't get how to use the `colorizer` API from `colors.js` to color the meta object output as gray (debug). Can you give me a pointer plaese

mcollina:
The `new chalk.Instance({ level })` API in https://github.com/pinojs/pino-pretty/blob/657b7ba081d1d8d8cc9fdb90d51c84a75c88d5f8/lib/colors.js#L18 will print out colors even if it's disabled in the TTY.

You will need to use `colorizer` with the `level` argument of `'trace'` (which is grey). The tricky part is that you'd need to percolate down the current colorizer from the caller function

lib/utils.js
```diff
@@ -3,6 +3,7 @@
 const dateformat = require('dateformat')
 const stringifySafe = require('fast-safe-stringify')
 const defaultColorizer = require('./colors')()
+const chalk = require('chalk')
```
jsumners, line 6: As Matteo said, we have a standardized colorizer in this module. It is defined on line 5 immediately before this new import. Please use that colorizer to add color to output.

gr2m:
Sorry I've spent another 15minutes on this I just don't know what to do with the `colorizer` argument. `colorizer` has a single method `.message()` which does not accept a level parameter. And calling colorizer as a function only returns the name of the level, not a chalk-compatible instance, e.g. `colorizer('trace') // TRACE`

Do you want to finish this up yourself? I'll learn from your changes for the next time

jsumners:
> Sorry I've spent another 15minutes on this I just don't know what to do with the `colorizer` argument. `colorizer` has a single method `.message()` which does not accept a level parameter. And calling colorizer as a function only returns the name of the level, not a chalk-compatible instance, e.g. `colorizer('trace') // TRACE`
> 
> Do you want to finish this up yourself? I'll learn from your changes for the next time

https://runkit.com/embed/op9ez5z3pgxm

`lib/colors` returns a factory function that asks "do you want colors or not?". In either case, it returns a new function that will return a colorized level label based upon the level given, e.g. `colorize('trace')` or `colorize('info')`. This function _also_ includes a `message` method that will colorize a message according to our standardized message color: `colorizer.message('a message')`.

What you want to do is add a new method that will colorize according to a different color, e.g. `colorizer.greyMessage = (msg) => colorizeLevel('trace', msg)`.

gr2m:
> What you want to do is add a new method that will colorize according to a different color, e.g. `colorizer.greyMessage = (msg) => colorizeLevel('trace', msg)`.

I would add that code to `lib/colors.js` as part of the factory function?

jsumners:
> I would add that code to `lib/colors.js` as part of the factory function?

I would expect to see it defined for each colorizer type in the same general location as the definition of `colorizer.message`.

gr2m:
see my commit at https://github.com/pinojs/pino-pretty/pull/161/commits/2d7ca36935d7cf574b39278d63023528c77992f4

That makes the tests fail locally now. How do I test for the color output?

jsumners:
> That makes the tests fail locally now. How do I test for the color output?

https://github.com/pinojs/pino-pretty/blob/657b7ba081d1d8d8cc9fdb90d51c84a75c88d5f8/test/lib/colors.test.js#L42

gr2m:
okay so I can write a unit test for the `colorizer.greyMessage` method, but I cannot test the color of the log output in basic.test.js?

jsumners:
> okay so I can write a unit test for the `colorizer.greyMessage` method, but I cannot test the color of the log output in basic.test.js?

Is there a need to? If the colorizer is working as tested, then the final result is going to be correct, yes?

gr2m:
okay, just wanted to make sure. I think it would be better to have some integration tests, because in theory the usage of `colorizer.greyMessage` in `prettifyObject()` could be removed, but the tests would continue to pass. 

I think I'm all set now, I'll finish this up. Thank you so much for bearing with me

gr2m:
I think this is ready

mcollina (approved):
lgtm

jsumners (approved):
LGTM

Should a note about the different coloring be added to the documentation for the single line option?

mcollina:
I don't think we should add any note.

Issue #132, opened 2020-09-16
Ignoring nested properties 


hugomallet:
Hello,

It looks like it is not possible to ignore nested properties.
It would be cool if we could ignore for example `--ignore req.headers,req.id` without ignoring completely the object req.

mcollina:
Would you like to send a Pull Request to address this issue? Remember to add unit tests.

acidoxee:
Hi, thanks for this update, unfortunately this seems to un-ignore keys that are composed of dots. For instance, we have a bunch of `logging.googleapis.com/xxxxxx` keys for GCP, which now can't be ignored.

jsumners:
@acidoxee please open a new issue and reference the PR that introduced the problem. This issue should have been closed with the PR merge.

PR #180, opened 2021-05-01
Add support to ignore nested properties

TzeWey:
This PR addresses #132 with a unit test to verify the change

### DIFF
--- a/index.js
+++ b/index.js
@@ -11,7 +11,8 @@ const {
   prettifyMessage,
   prettifyMetadata,
   prettifyObject,
-  prettifyTime
+  prettifyTime,
+  filterLog
 } = require('./lib/utils')
 
 const bourne = require('@hapi/bourne')
@@ -81,12 +82,7 @@ module.exports = function prettyFactory (options) {
     const prettifiedMessage = prettifyMessage({ log, messageKey, colorizer, messageFormat, levelLabel })
 
     if (ignoreKeys) {
-      log = Object.keys(log)
-        .filter(key => !ignoreKeys.has(key))
-        .reduce((res, key) => {
-          res[key] = log[key]
-          return res
-        }, {})
+      log = filterLog(log, ignoreKeys)
     }
 
     const prettifiedLevel = prettifyLevel({ log, colorizer, levelKey })
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -21,13 +21,15 @@ module.exports = {
   prettifyMessage,
   prettifyMetadata,
   prettifyObject,
-  prettifyTime
+  prettifyTime,
+  filterLog
 }
 
 module.exports.internals = {
   formatTime,
   joinLinesWithIndentation,
-  prettifyError
+  prettifyError,
+  deleteLogProperty
 }
 
 /**
@@ -434,4 +436,39 @@ function prettifyError ({ keyName, lines, eol, ident }) {
   }
 
   return result
+}
+
+/**
+ * Deletes a specified property from a log object if it exists
+ *
+ * @param {object} log The log object to be modified.
+ * @param {string} property A string identifying the property to be deleted from
+ * the log object. Accepts nested properties delimited by a `.`
+ * e.g. `'prop1.prop2'`.
+ */
+function deleteLogProperty (log, property) {
+  const props = property.split('.')
+  const propToDelete = props.pop()
+
+  props.forEach((prop) => {
+    if (!Object.prototype.hasOwnProperty.call(log, prop)) {
+      return
+    }
+    log = log[prop]
+  })
+
+  delete log[propToDelete]
+}
+
+/**
+ * Filter a log object by removing any ignored keys.
+ *
+ * @param {object} log The log object to be modified.
+ * @param {string} ignoreKeys An array of strings identifying the properties to be removed.
+ */
+function filterLog (log, ignoreKeys) {
+  ignoreKeys.forEach((ignoreKey) => {
+    deleteLogProperty(log, ignoreKey)
+  })
+  return log
 }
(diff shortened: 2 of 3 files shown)


index.js
```diff
@@ -81,12 +82,7 @@ module.exports = function prettyFactory (options) {
     const prettifiedMessage = prettifyMessage({ log, messageKey, colorizer, messageFormat, levelLabel })
 
     if (ignoreKeys) {
-      log = Object.keys(log)
-        .filter(key => !ignoreKeys.has(key))
-        .reduce((res, key) => {
-          res[key] = log[key]
-          return res
-        }, {})
+      log = filterLog(log, ignoreKeys)
```
mcollina, line 85: Can you avoid calling this if there are no ignoreKeys?

lib/utils.js
```diff
@@ -435,3 +437,38 @@ function prettifyError ({ keyName, lines, eol, ident }) {
...
+  const propToDelete = props.pop()
+
+  props.forEach((prop) => {
+    if (!Object.prototype.hasOwnProperty.call(log, prop)) {
+      return
+    }
+    log = log[prop]
+  })
+
+  delete log[propToDelete]
```
mcollina, line 460: the original implementation avoided modifying the existing object. I prefer allocating a new object in this case.

mcollina (changes requested):
Good work! I've left a few comments

index.js
TzeWey: Is the check at line 84 sufficient? Or do you prefer an explicit length check as well?

lib/utils.js
TzeWey: I did think about this for a while but was not able to come up with a solution.
As the ignored keys can now be nested, we would need to perform a deep clone of the log object to avoid modifying the existing object. 
Is there a performance concern with doing a deep clone? Or if you have some other method in mind?
May I know what is the concern with modifying the original object? Thanks!

index.js
mcollina: Oh I missed it! Great!

lib/utils.js
mcollina: I agree, do a deep clone before filtering. You can use `rdfc` for this.

TzeWey: I assume it was a typo and you actually meant `rfdc`.
In the latest commit, the package was added and is used to clone the log object before filtering.

index.js
TzeWey: Okie, I shall leave this as it is. Thanks!

mcollina (commented):
Can you add docs for this feature?

TzeWey:
Updated README

mcollina (approved):
lgtm

lib/utils.js
```diff
@@ -435,3 +438,39 @@ function prettifyError ({ keyName, lines, eol, ident }) {
 
   return result
 }
+
+/**
+ * Deletes a specified property from a log object if it exists
```
jsumners, line 443:
```suggestion
 * Deletes a specified property from a log object if it exists.
 * This function mutates the passed in `log` object.
```

lib/utils.js
```diff
@@ -435,3 +438,39 @@ function prettifyError ({ keyName, lines, eol, ident }) {
...
+  })
+
+  delete log[propToDelete]
+}
+
+/**
+ * Filter a log object by removing any ignored keys.
+ *
+ * @param {object} log The log object to be modified.
+ * @param {string} ignoreKeys An array of strings identifying the properties to be removed.
```
jsumners, line 468:
```suggestion
 * @param {string} ignoreKeys An array of strings identifying the properties to be removed.
 *
 * @returns {object} A new `log` object instance that does not include the ignored keys.
```

lib/utils.js
```diff
@@ -435,3 +438,39 @@ function prettifyError ({ keyName, lines, eol, ident }) {
...
+
+/**
+ * Deletes a specified property from a log object if it exists
+ *
+ * @param {object} log The log object to be modified.
+ * @param {string} property A string identifying the property to be deleted from
+ * the log object. Accepts nested properties delimited by a `.`
+ * e.g. `'prop1.prop2'`.
+ */
+function deleteLogProperty (log, property) {
```
jsumners, line 450: Please add tests to https://github.com/pinojs/pino-pretty/blob/d9fc678fb546c466e9df02cd8eb4da35a8c26e7e/test/lib/utils.internals.test.js

lib/utils.js
```diff
@@ -435,3 +438,39 @@ function prettifyError ({ keyName, lines, eol, ident }) {
...
+  delete log[propToDelete]
+}
+
+/**
+ * Filter a log object by removing any ignored keys.
+ *
+ * @param {object} log The log object to be modified.
+ * @param {string} ignoreKeys An array of strings identifying the properties to be removed.
+ */
+function filterLog (log, ignoreKeys) {
```
jsumners, line 470: Please add tests to https://github.com/pinojs/pino-pretty/blob/d9fc678fb546c466e9df02cd8eb4da35a8c26e7e/test/lib/utils.public.test.js

jsumners (changes requested):

lib/utils.js
TzeWey: Updated in latest commit

lib/utils.js
TzeWey: Updated in latest commit

lib/utils.js
TzeWey: Updated in the latest commit

lib/utils.js
TzeWey: Added in the latest commit

jsumners (approved):

PR #183, opened 2021-05-06
chore: documentation and CI updates

Fdawgs:


mcollina (approved):
lgtm

Fdawgs:
@mcollina, mind if i open similar requests on other pino repos?
Some don't have CI in place (or are on travis still), so wanted to fix that!

mcollina:
go for it!

package.json
```diff
@@ -7,9 +7,9 @@
     "pino-pretty": "./bin.js"
   },
   "scripts": {
-    "ci": "standard && tap 'test/**/*.test.js' --coverage-report=lcovonly",
+    "ci": "standard && tap \"test/**/*.test.js\" --coverage-report=lcovonly",
     "lint": "standard | snazzy",
-    "test": "tap --100 'test/**/*.test.js'"
+    "test": "tap --100 \"test/**/*.test.js\""
```
jsumners, line 12: These changes should not have been made. The `"` will introduce shell expansion of `**/*.test.js`. The `'` prevents shell expansion and allows the `glob` package to do the expansion.

@Fdawgs please do not blanket change quotes.

Fdawgs: Ah, sorry @jsumners. Is there a better way of doing this?

The Fastify repos all use a similar method to ensure the tests run on Windows.

jsumners: They should all be using single quotes.

Fdawgs: After further investigation, I don't think it matters if single or double quotes are used when it's just globbing (phew, no mass opening of PRs on all of Fastify's repos):

```bash
$ cd test

$ echo *
decorate.test.js hook.test.js integration.test.js plugin.test-d.ts

$ echo "*"
*

$ echo '*'
*
```

PR #186, opened 2021-05-19
Add TypeScript types

kibertoad:
Part of effort to do https://github.com/pinojs/pino/pull/913

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,110 @@
+// Type definitions for pino-pretty 4.7
+// Project: https://github.com/pinojs/pino-pretty#readme
+// Definitions by: Adam Vigneaux <https://github.com/AdamVig>
+// Minimum TypeScript Version: 3.0
+
+type LogDescriptor = Record<string, unknown>;
+
+declare function PinoPretty(options: PinoPretty.PrettyOptions): PinoPretty.Prettifier;
+
+declare namespace PinoPretty {
+  type Prettifier = (inputData: string | object) => string;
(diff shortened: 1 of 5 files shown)


mcollina (approved):
lgtm

squidini:
This kind of breaks pino with `TS2694: Namespace '"/node_modules/pino-pretty/index"' has no exported member 'PrettyOptions'.` #189

kibertoad:
@squidini Thank you for the bug report! Will fix this today

kibertoad:
@squidini Isn't it actually available on a namespace, though?

squidini:
@kibertoad it actually is available in a namespace.
But defining namespace this way breaks support of `esnext` modules.
I managed to resolve this issue by declaring everything in a module.

```ts
declare module "pino-pretty" {
  export interface PrettyOptions {
```

This is my tsconfig:

```json
{
  "compilerOptions": {
    "target": "ES2020",
    "module": "esnext",
    "moduleResolution": "node",
    "esModuleInterop": true,
    "strict": true,
  }
}
```

kibertoad:
Wonder why it got broken, though. Wasn't it not exported directly in the past either? See https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/pino-pretty/index.d.ts

squidini:
Probably it wasn't working with es2020 modules before. I need to verify this.

kibertoad:
@squidini Would appreciate it! Will still try to make it work with both options, though.

PR #194, opened 2021-06-01
updated to tap@15

mcollina:
to support npm@7


jsumners (approved):
Looks good to me.

No idea what `npm@7` has to do with it though.

PR #47, opened 2018-11-27
Align log levels

Zirak:
INFO and WARN lines are now properly aligned to their brethren, making
it easier to scan log lines for all the great people on earth and
elsewhere.

Example script:

```js
const logger = require('pino')({
  prettyPrint: true,
  prettifier: require('.')
})

logger.info('Ini')
logger.debug('Mini')
logger.error('Miny')
logger.warn('Mo!')
```

Before:
```js
[1543354160314] INFO (15120 on sputnik): Ini
[1543354160316] DEBUG (15120 on sputnik): Mini
[1543354160316] ERROR (15120 on sputnik): Miny
[1543354160316] WARN (15120 on sputnik): Mo!
```

After:
```js
[1543354145360] INFO  (12712 on sputnik): Ini
[1543354145362] DEBUG (12712 on sputnik): Mini
[1543354145362] ERROR (12712 on sputnik): Miny
[1543354145363] WARN  (12712 on sputnik): Mo!
```

What a crazy crazy world we live in.

PR #197, opened 2021-06-07
feat: align log levels

Tchoupinax:
Hello,

It's been several times that I find it unfortunate that the logs are not aligned. I searched how to align logs and I found this PR #47 (3 years ago) that fixes the issue but this fix has disappeared from master.

The current PR fixes the issue again :) It is more beautiful with ^^

Thank you for your work!

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -21,7 +21,7 @@ A standard Pino log line like:
 Will format to:
 
 ```
-[1522431328992] INFO (42 on foo): hello world
+[1522431328992] INFO  (42 on foo): hello world
 ```
 
 <a id="example"></a>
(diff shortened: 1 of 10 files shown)


mcollina:
I don't think this is doable as we had to revert because it created more work than necessary.

wdyt @jsumners?

jsumners:
Thank you for your contribution, but this feature introduces more problems than it solves. This feature makes maintenance very difficult. We cannot accept this feature.

Tchoupinax:
Thank you for your answer.

Could you enum exemples of point that makes maintenance difficult ? (For my information and maybe have a trace for another people)

jsumners:
Just search back through the issues and pull requests for the ones around alignment.

PR #209, opened 2021-07-19
Fix 'ignore' for keys containing key delimiter

TzeWey:
This PR fixes the issue reported by @acidoxee in https://github.com/pinojs/pino-pretty/issues/132#issuecomment-861608098 just before the issue was closed. 

This commit attempts to overcome the limitation by allowing the user to escape the dot character used as a delimiter.

The method used is to split the key by the dot delimiter ignoring any escaped dots, followed by replacing all occurrences of the escaped dots. Alternatively, it is possible to use a negative lookbehind regex for the split, but might not be supported on all browsers.

Thanks!

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -31,6 +31,7 @@ module.exports.internals = {
   joinLinesWithIndentation,
   prettifyError,
   deleteLogProperty,
+  splitIgnoreKey,
   createDate,
   isValidDate
 }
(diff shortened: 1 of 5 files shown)


lib/utils.js
```diff
@@ -486,10 +486,11 @@ function prettifyError ({ keyName, lines, eol, ident }) {
  * @param {object} log The log object to be modified.
  * @param {string} property A string identifying the property to be deleted from
  * the log object. Accepts nested properties delimited by a `.`
- * e.g. `'prop1.prop2'`.
+ * Delimiter can be escaped to preserve property names that contain the delimiter.
+ * e.g. `'prop1.prop2'` or `'prop2\.domain\.com.prop2'`
  */
 function deleteLogProperty (log, property) {
-  const props = property.split('.')
+  const props = property.match(/([^\\.]|\\\.)+/g).map((e) => e.split('\\.').join('.'))
```
kibertoad, line 493: how expensive is this operation?

mcollina (approved):
lgtm

lib/utils.js
TzeWey: https://jsbench.me/tqkrafwdkf/1

I just performed some benchmarks, this implementation is ~81% slower compared to just a split.

There might be quite a hit on performance with this change, however, this can be improved if we split the key using a loop. Benchmark shows that it is ~36% slower compared to just a split.

Will be pushing the improved version shortly for review.

kibertoad: Thank you!

acidoxee:
Thanks very much for this PR!

PR #216, opened 2021-08-08
Fix shortcut usage of '-a' and '-m' options

### LINKED ISSUE #212: shortcut -a won't work
Hi, I'm using pino-pretty on another logger, so I need to change the timestamp key like this:

```
# expected ts to be timestamp key
$> echo '{"ts":1623170292956,"msg":"one"}' | pino-pretty -a ts
one
    ts: 1623170292956
```

But it won't translate it correctly. After trial and error I found out, that the long version will do the trick:

```
# works
$> echo '{"ts":1623170292956,"msg":"one"}' | pino-pretty --timestampKey ts
[1623170292956]: one
```

Looks like the shortcut `-a` isn't working correctly.

micalevisk:
Fixes #212

---

I'm not sure if the tests could be improved.

### DIFF
--- a/bin.js
+++ b/bin.js
@@ -70,6 +70,12 @@ let opts = args.parse(process.argv, {
       messageKey: DEFAULT_VALUE,
       levelKey: DEFAULT_VALUE,
       timestampKey: DEFAULT_VALUE
+    },
+    // NOTE: The following key-value pairs values should be in sync with the
+    //       short version values defined in each `args.option([value, key], ...)`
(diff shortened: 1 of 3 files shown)


mcollina (changes requested):
Could you create multiple tests instead of looping within a test? Looping _outside_ of a test is ok too.

mcollina (approved):
lgtm

jsumners (approved):
Looks good to me.

PR #222, opened 2021-08-29
Use pino-abstract-transport

mcollina:
This change makes this module compatible with the new transport system.

This is the first step for https://github.com/pinojs/pino/issues/1106.

It is a semver-major change and it will require a minor change in pino@6 and pino@7 to achieve compatibility. before we exported the `prettyFactory`, now we export the function to build a transport, which includes split2.

### DIFF
--- a/index.js
+++ b/index.js
@@ -5,8 +5,10 @@ const pump = require('pump')
 const { Transform } = require('readable-stream')
 const abstractTransport = require('pino-abstract-transport')
 const jmespath = require('jmespath')
-const colors = require('./lib/colors')
 const sonic = require('sonic-boom')
+const bourne = require('@hapi/bourne')
+
+const colors = require('./lib/colors')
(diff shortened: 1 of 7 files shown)


mcollina:
@jsumners I know you preferred this to not become compatible with the new system, but the changes are minimal so I thought it was worth a shot.

mcollina:
cc @leorossi

index.js
```diff
@@ -165,6 +169,41 @@ function prettyFactory (options) {
...
+    })
+
+    pump(source, stream, destination)
+    return stream
+  }, { parse: 'lines' })
+}
+
+module.exports = build
+module.exports.prettyFactory = prettyFactory
+module.exports.default = build
```
jsumners, line 209: This is semver major.

mcollina:
@jsumners updated the require order.

jsumners (approved):

Issue #223, opened 2021-08-31
jmespath uses an old, vulnerable version of jQuery


BitForger:
`jmespath` is currently using jQuery 1.7.2 which is referenced specifically by [this vulnerability](https://nvd.nist.gov/vuln/detail/CVE-2012-6708) in 2012. Is it possible to move to a different utility for this? It appears jmespath is unmaintained.

kibertoad:
PR would be welcome :)

jsumners:
How is that CVE applicable to a local CLI tool?

BitForger:
> How is that CVE applicable to a local CLI tool?

I'll preface with I'm not a hacker or pen-tester but from my opinion...

I agree that the vulnerability in the context of this code is basically void, but it does appear in a dependency graph and nearly every tool that seeks out and flags vulnerabilities is flagging that dependency.

This is mostly a paperwork issue for enterprises who are trying to make sure attack vectors are mitigated. If you are open to finding a suitable replacement I can certainly look for and create a PR for it.

jsumners:
We are certainly open.

mcollina:
I think a replacement could be https://jsonata.org/. Wdyt @jsumners?

jsumners:
I haven't used jmespath in a long time, but I know it's used by Ansible. JSONata would be a complete query language change and result in a semver major requiring documentation updates and a clear warning with comparitive example in the release notes. 

I am okay with whatever works.

BitForger:
Here is a [fork and TS implementation](https://github.com/nanoporetech/jmespath-ts) that might be usable. Other than that, is jsonpath the more standard one being used? I've seen mostly jsonpath and jq being used primarily in my personal experience.

BitForger:
Sure. I'll start work on it tonight

tilleps:
How is jmespath using jQuery?  I don't see any references to jQuery:
https://github.com/jmespath/jmespath.js/search?q=jquery
https://github.com/jmespath/jmespath.js/blob/master/package.json

BitForger:
I believe it might be grunt has a dependency on it. I'm laid up sick right now, but I'll try to check it tonight.

mcollina:
@BitForger jamespath has no runtime dependencies at all.

BitForger:
[...]

PR #224, opened 2021-09-02
feat: remove jmespath and replace with jsonpath

BitForger:
closes pinojs/pino-pretty#223


mcollina (approved):
lgtm

mcollina:
Given this is a semver-major change, it will likely ship with https://github.com/pinojs/pino-pretty/pull/222

jsumners (commented):
I don't think I agree with this. JSONPath is more of an access language while JMESPath implements a full query language. Maybe we should fork and update the original library?

PR #230, opened 2021-09-17
Update colorette to enable behaviour consistent with earlier pino-pretty versions

### LINKED ISSUE #220: No color on node 12 server
Noticed after upgrading to pino-pretty 6.0.0 so some problem was probably introduced by colorette in PR #218 

In my utils I have:
```
import pino from "pino";
import prettifier from "pino-pretty";

export const getLogger = (level = "warn"): pino.Logger => pino({
  level,
  prettyPrint: {
    colorize: true,
    ignore: "pid,hostname,module",
    messageFormat: `[{module}] {msg}`,
    translateTime: true,
  },
  prettifier,
});
```

Then on my server I have:
```
import { getLogger } from "@valuemachine/utils";

import { env } from "./env";

const log = getLogger(env.logLevel).child({ module: "Utils" });
log.info(`Logger loaded!`);
```

Result is white text:
Screenshot from 2021-08-24 13-53-14

Downgrading to pino-pretty 5.1.3 results in colored output:
Screenshot from 2021-08-24 13-55-28

This server is based on the `node:12.16.1-alpine3.11` docker image so it shouldn't be affected by colorette's problems w node v13

kibertoad:
fixes #220

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -132,7 +132,7 @@ with keys corresponding to the options described in [CLI Arguments](#cliargs):
 
 ```js
 {
-  colorize: colorette.options.enabled, // --colorize
+  colorize: colorette.isColorSupported, // --colorize
   crlf: false, // --crlf
   errorLikeObjectKeys: ['err', 'error'], // --errorLikeObjectKeys
   errorProps: '', // --errorProps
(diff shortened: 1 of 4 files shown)


mcollina (approved):
lgtm

PR #234, opened 2021-09-21
fix: dateformat migrated to esm

climba03003:
As `dateformat` moved from `CJS` to `ESM`.
Change the version to `~` flag to prevent further impact.

Related to https://github.com/pinojs/pino/pull/1137

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,6 +13,8 @@ on:
 jobs:
   build:
     runs-on: ${{ matrix.os }}
+    outputs:
+      COVERALLS: ${{ steps.coveralls-trigger.outputs.COVERALLS_TRIGGER }}
     strategy:
       matrix:
         node-version: [10, 12, 14, 16]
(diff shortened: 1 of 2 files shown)


climba03003:
It seems that coveralls is blocking the CI. `405` problems again.

kibertoad:
Should be finished soon: https://status.coveralls.io/?utm_source=embed

climba03003:
I will either use [`continue-on-error`](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error) for `Coveralls` or disable it until the maintenance is over.

Coveralls should not be the blocker for CI.

kibertoad:
continue-on-error sounds like a great idea!

kibertoad:
@climba03003 Can you try 4.6.2? If I'm reading https://github.com/felixge/node-dateformat/releases right, looks like they realized their mistake.

climba03003:
> 
> 
> @climba03003 Can you try 4.6.2? If I'm reading https://github.com/felixge/node-dateformat/releases right, looks like they realized their mistake.

@kibertoad I will use 4.6.3 as it is the latest in `npmjs`. And I have updated the CI to skip or allow all the stage of coveralls to fail.

mcollina (approved):
lgtm

PR #235, opened 2021-09-21
Switch to @matteo.collina/dateformat

mcollina:
The dateformat maintainers decided that the v5.0.0 release would be
ESM-only. Give the current stance of keeping the pino ecosystem to CJS
for the time being, I do not think it is a safe long-term to depend
on this library. Therefore I forked it.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -1,7 +1,7 @@
 'use strict'
 
 const clone = require('rfdc')({ circles: true })
-const dateformat = require('dateformat')
+const dateformat = require('@matteo.collina/dateformat')
 const stringifySafe = require('fast-safe-stringify')
 const defaultColorizer = require('./colors')()
 const {
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
   "dependencies": {
     "args": "^5.0.1",
     "colorette": "^2.0.0",
-    "dateformat": "^4.5.1",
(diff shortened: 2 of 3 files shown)


kibertoad:
What is the benefit over just pinning to 4.x?

mcollina:
The benefit is that this new library could actually be maintained and not pinned in the past.

kibertoad (approved):

mcollina:
Note that pinning to v4 only is correct and will get the job done for the time being.

jsumners:
I'm concerned about tracking upstream maintenance. This is a very seldomly updated module. Think it will be difficult to realize improvements have been made and to port them to the fork. Can we open an issue upstream to ask them to reconsider?

kibertoad:
@jsumners They already did proper semver and pushed the change to a new 5.x branch. Considering the uptake in ESM usage lately, I don't think asking them to consider dropping ESM entirely is a fair ask.

kibertoad:
At best we can send PR for integrating one of the dual publishing solutions, I guess.

mcollina:
Posted https://github.com/felixge/node-dateformat/issues/169

jsumners:
> Posted [felixge/node-dateformat#169](https://github.com/felixge/node-dateformat/issues/169)

Thanks. Let's see what outcome we get before proceeding with this path.

PR #245, opened 2021-10-14
Updated to pino@7

mcollina:
This is an update to pino@7.
The main difference with handling errors is that now pino uses `err` property by default.

I retained the support of the previous (pino@6) syntax as well so it still works with pino@6.

I do not plan to bump the major unless anybody prefers otherwise (pino-pretty@7 was released to match pino@7)

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -187,6 +187,7 @@ function prettifyErrorLog ({
       // Print all sibling properties except for the standard exclusions.
       propertiesToPrint = Object.keys(log).filter(k => excludeProperties.includes(k) === false)
     } else {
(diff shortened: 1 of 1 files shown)


lib/utils.js
```diff
@@ -187,7 +187,6 @@ function prettifyErrorLog ({
       // Print all sibling properties except for the standard exclusions.
       propertiesToPrint = Object.keys(log).filter(k => excludeProperties.includes(k) === false)
     } else {
-      // Print only specified properties unless the property is a standard exclusion.
```
jsumners, line 190: Why remove this comment?

mcollina: my bad, it was not intended.

PR #256, opened 2021-11-01
Support SonicBoom as a destination

mcollina:
Adds a couple of missing options

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -211,6 +211,7 @@ The options accepted have keys corresponding to the options described in [CLI Ar
   // The file or file descriptor (1 is stdout) to write to
   destination: 1,
 
+  // Alternatively, pass a `sonic-boom` instance (allowing more flexibility):
   // destination: new SonicBoom({ dest: 'a/file', mkdir: true })
 
   customPrettifiers: {}
--- a/Readme.md
+++ b/Readme.md
@@ -207,6 +207,11 @@ The options accepted have keys corresponding to the options described in [CLI Ar
   ignore: 'pid,hostname', // --ignore
   hideObject: false, // --hideObject
   singleLine: false, // --singleLine
+
+  dest: 1, // The file or file descriptor to write to
+  append: true, // Append to the destination
+  mkdir: false, // Ensure the directory for dest exist when true
+
   customPrettifiers: {}
 }
 ```
--- a/Readme.md
+++ b/Readme.md
@@ -208,9 +208,10 @@ The options accepted have keys corresponding to the options described in [CLI Ar
   hideObject: false, // --hideObject
   singleLine: false, // --singleLine
 
-  dest: 1, // The file or file descriptor to write to
-  append: true, // Append to the destination
-  mkdir: false, // Ensure the directory for dest exist when true
+  // The file or file descriptor (1 is stdout) to write to
+  destination: 1,
+
+  // destination: new SonicBoom({ dest: 'a/file', mkdir: true })
 
   customPrettifiers: {}
 }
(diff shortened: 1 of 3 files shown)


Readme.md
```diff
@@ -207,6 +207,11 @@ The options accepted have keys corresponding to the options described in [CLI Ar
   ignore: 'pid,hostname', // --ignore
   hideObject: false, // --hideObject
   singleLine: false, // --singleLine
+
+  dest: 1, // The file or file descriptor to write to
+  append: true, // Append to the destination
```
kibertoad, line 212: What happens if it is false? Will it overwrite the destination?

Readme.md
```diff
@@ -207,6 +207,11 @@ The options accepted have keys corresponding to the options described in [CLI Ar
...
+  mkdir: false, // Ensure the directory for dest exist when true
```
kibertoad, line 213: Does this work recursively?

Readme.md
```diff
@@ -207,6 +207,11 @@ The options accepted have keys corresponding to the options described in [CLI Ar
   ignore: 'pid,hostname', // --ignore
   hideObject: false, // --hideObject
   singleLine: false, // --singleLine
+
+  dest: 1, // The file or file descriptor to write to
```
kibertoad, line 211: what does default destination of 1 mean?

test/basic.test.js
```diff
@@ -5,9 +5,16 @@ const os = require('os')
 const test = require('tap').test
 const pino = require('pino')
 const dateformat = require('dateformat')
+const path = require('path')
+const rimraf = require('rimraf')
```
kibertoad, line 9: Since we only support Node 12+, rimraf is redundant, you can use `fs.rmSync(dir, { recursive: true, force: true });`

Readme.md
mcollina: Yes

Readme.md
mcollina: Yes

Readme.md
mcollina: the file descriptor of stdout.

test/basic.test.js
mcollina: I'm not doing that because it changed again in v16, so this move it a bit more far out in the future :(.

index.js
```diff
@@ -183,7 +183,12 @@ function build (opts = {}) {
       }
     })
 
-    const destination = sonic({ dest: opts.destination || 1, sync: false })
+    const destination = sonic({
```
jsumners, line 186: Could the user pass in a `SonicBoom` instance instead? This change leads us down the path of having to maintain 1:1 compatibility with all of the `SonicBoom` constructor options.

Leave the current default, but check for `opts.destination is SonicBoom` before it. That will make it a backward compatible change.

jsumners (changes requested):

mcollina:
I've updated it, let me know what you think

kibertoad (approved):

Readme.md
```diff
@@ -207,6 +207,12 @@ The options accepted have keys corresponding to the options described in [CLI Ar
   ignore: 'pid,hostname', // --ignore
   hideObject: false, // --hideObject
   singleLine: false, // --singleLine
+
+  // The file or file descriptor (1 is stdout) to write to
+  destination: 1,
+
+  // destination: new SonicBoom({ dest: 'a/file', mkdir: true })
```
jsumners, line 214:
```suggestion
  // The file or file descriptor (1 is stdout) to write to
  destination: 1,

  // Alternatively, pass a `sonic-boom` instance (allowing more flexibility):
  // destination: new SonicBoom({ dest: 'a/file', mkdir: true })
```

jsumners (approved):
Looks good to me.

PR #261, opened 2021-11-08
Update to sonic-boom@2.2.1

aaestrada:
This pr will Update to sonic-boom@2.2.1

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
     "readable-stream": "^3.6.0",
     "rfdc": "^1.3.0",
     "secure-json-parse": "^2.4.0",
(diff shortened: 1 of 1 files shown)


aaestrada:
Let me know if there is any request changes!

kibertoad:
What is the rationale behind this change?

aaestrada:
> What is the rationale behind this change?



The pino library is using ` "sonic-boom": "^2.2.1",` as a [dependency in the package.json](https://github.com/pinojs/pino/blob/master/package.json#L113)

so this will work better with the same version!

mcollina:
npm will resolve to be the same version when it installs.

Issue #185, opened 2021-05-11
Seems to be no way to remove the brackets of the timestamp?


airship-shrimp:
Can I achieve it in some way?

jsumners:
Not with a PR to implement such a feature.

https://github.com/pinojs/pino-pretty/blob/7693cd4810ab7b9589f33364148d0cd33119dd42/lib/utils.js#L390-L405

pearofducks:
@airship-shrimp - once 7.3.0 (assuming this is the version) is released, you'll be able to use `customPrettifiers.time` to customize this and remove the brackets.

Fdawgs:
closed by #263

PR #263, opened 2021-11-14
Add ability to format time and level with user-defined prettifiers

pearofducks:
This would close issues like #185, and put more formatting capabilities into userland.

If another option-key besides `customPrettifiers` should be used I'm happy to switch. I think the docs are a bit open to interpretation even though I see this option is only being used to prettify object-formatting right now.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -238,7 +238,8 @@ const prettifyQuery = value => {
 }
 ```
 
-Additionally, `customPrettifiers` can be used to format the time and level outputs:
+Additionally, `customPrettifiers` can be used to format the time and level
+outputs:
 ```js
 {
   customPrettifiers: {
@@ -252,18 +253,6 @@ Additionally, `customPrettifiers` can be used to format the time and level outpu
     level: logLevel => `LEVEL: ${logLevel}`
   }
 }
-```
-
-Note that prettifiers do not include any coloring, if the stock coloring on `level` is desired, it can be accomplished using the following:
-```js
-const { colorizerFactory } = require('pino-pretty')
-const levelColorize = colorizerFactory(true)
-const levelPrettifier = logLevel => `LEVEL: ${levelColorize(logLevel)}`
-//...
-{
-  customPrettifiers: { level: levelPrettifier }
-}
-```
 
 `messageFormat` option allows you to customize the message output. A template `string` like this can define the format:
 ```js
--- a/Readme.md
+++ b/Readme.md
@@ -238,21 +238,20 @@ const prettifyQuery = value => {
 }
 ```
 
-You can also use `customPrettifiers` to format the time and level outputs:
+Additionally, `customPrettifiers` can be used to format the time and level outputs:
 ```js
 {
   customPrettifiers: {
-    // the argument for this function will be the same
-    // string that's at the start of the log-line by default
+    // The argument for this function will be the same
+    // string that's at the start of the log-line by default:
     time: timestamp => `🕰 ${timestamp}`,
-    // the argument for the level-prettifier may vary depending
-    // on if you're using the levelKey option or not.
-    // by default this will be the same numerics as the Pino default
+    
+    // The argument for the level-prettifier may vary depending
+    // on if the levelKey option is used or not.
+    // By default this will be the same numerics as the Pino default:
     level: logLevel => `LEVEL: ${logLevel}`
   }
 }
-```
-
 `messageFormat` option allows you to customize the message output. A template `string` like this can define the format:
 ```js
 {
--- a/Readme.md
+++ b/Readme.md
@@ -254,6 +254,17 @@ outputs:
   }
 }
 
+Note that prettifiers do not include any coloring, if the stock coloring on
+`level` is desired, it can be accomplished using the following:
+```js
+const { colorizerFactory } = require('pino-pretty')
+const levelColorize = colorizerFactory(true)
+const levelPrettifier = logLevel => `LEVEL: ${levelColorize(logLevel)}`
+//...
+{
+  customPrettifiers: { level: levelPrettifier }
+}
+
 `messageFormat` option allows you to customize the message output. A template `string` like this can define the format:
 ```js
 {
--- a/Readme.md
+++ b/Readme.md
@@ -256,7 +256,8 @@ Additionally, `customPrettifiers` can be used to format the time and level outpu
 
 Note that prettifiers do not include any coloring, if the stock coloring on `level` is desired, it can be accomplished using the following:
 ```js
-const levelColorize = require('pino-pretty/lib/colors.js')(true)
+const { colorizerFactory } = require('pino-pretty')
+const levelColorize = colorizerFactory(true)
 const levelPrettifier = logLevel => `LEVEL: ${levelColorize(logLevel)}`
 //...
 {
--- a/Readme.md
+++ b/Readme.md
@@ -245,13 +245,25 @@ Additionally, `customPrettifiers` can be used to format the time and level outpu
     // The argument for this function will be the same
     // string that's at the start of the log-line by default:
     time: timestamp => `🕰 ${timestamp}`,
-    
+
     // The argument for the level-prettifier may vary depending
     // on if the levelKey option is used or not.
     // By default this will be the same numerics as the Pino default:
     level: logLevel => `LEVEL: ${logLevel}`
   }
 }
+```
+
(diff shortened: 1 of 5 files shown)


mcollina (commented):
Could you document this option?

lib/utils.js
```diff
@@ -223,9 +223,10 @@ function prettifyErrorLog ({
  * `undefined` will be returned. Otherwise, a string from the specified
  * `colorizer` is returned.
  */
-function prettifyLevel ({ log, colorizer = defaultColorizer, levelKey = LEVEL_KEY }) {
+function prettifyLevel ({ log, colorizer = defaultColorizer, levelKey = LEVEL_KEY, prettifier }) {
```
jsumners, line 226: Missing documentation in the docblock.

lib/utils.js
```diff
@@ -428,7 +429,7 @@ function prettifyObject ({
  * `undefined` is returned. Otherwise, the prettified time is returned as a
  * string.
  */
-function prettifyTime ({ log, timestampKey = TIMESTAMP_KEY, translateFormat = undefined }) {
+function prettifyTime ({ log, timestampKey = TIMESTAMP_KEY, translateFormat = undefined, prettifier }) {
```
jsumners, line 432: Missing documentation in the docblock.

lib/utils.js
```diff
@@ -223,9 +223,10 @@ function prettifyErrorLog ({
  * `undefined` will be returned. Otherwise, a string from the specified
  * `colorizer` is returned.
  */
-function prettifyLevel ({ log, colorizer = defaultColorizer, levelKey = LEVEL_KEY }) {
+function prettifyLevel ({ log, colorizer = defaultColorizer, levelKey = LEVEL_KEY, prettifier }) {
   if (levelKey in log === false) return undefined
-  return colorizer(log[levelKey])
+  const output = log[levelKey]
+  return prettifier ? prettifier(output) : colorizer(output)
```
jsumners, line 229: Does this not break colorization?

jsumners (changes requested):

lib/utils.js
pearofducks: Do you mean colorization via the user-supplied `prettifier` function? My intent is that the user owns any/all formatting (including colorizing) via this function.

I personally want to customize the coloring done via this method to alter it from the default supplied here.

If this is a concern I'm happy to document how to obtain the built-in behavior in the prettifier so users who only wish to lightly customize output can do so.

jsumners: > If this is a concern I'm happy to document how to obtain the built-in behavior in the prettifier so users who only wish to lightly customize output can do so.

That would be acceptable.

Readme.md
```diff
@@ -238,6 +238,21 @@ const prettifyQuery = value => {
...
+    // string that's at the start of the log-line by default
+    time: timestamp => `🕰 ${timestamp}`,
+    // the argument for the level-prettifier may vary depending
+    // on if you're using the levelKey option or not.
+    // by default this will be the same numerics as the Pino default
+    level: logLevel => `LEVEL: ${logLevel}`
+  }
+}
+```
+
```
jsumners, line 255:
```suggestion
Additionally, `customPrettifiers` can be used to format the time and level outputs:
```js
{
  customPrettifiers: {
    // The argument for this function will be the same
    // string that's at the start of the log-line by default:
    time: timestamp => `🕰 ${timestamp}`,
    
    // The argument for the level-prettifier may vary depending
    // on if the levelKey option is used or not.
    // By default this will be the same numerics as the Pino default:
    level: logLevel => `LEVEL: ${logLevel}`
  }
}
```

```

File: lib/utils.js [hunk 5]
```diff
@@ -218,14 +218,16 @@ function prettifyErrorLog ({
  * @param {function} [input.colorizer] A colorizer function that accepts a level
  * value and returns a colorized string. Default: a no-op colorizer.
  * @param {string} [levelKey='level'] The key to find the level under.
+ * @param {function} input.colorizer A user-supplied formatter to be called instead of colorizer
```
jsumners, line 221:
```suggestion
 * @param {function} [input.prettifier] A user-supplied formatter to be called instead of colorizer.
```

lib/utils.js
```diff
@@ -423,12 +425,13 @@ function prettifyObject ({
  * timestamp will be prettified into a string at UTC using the default
  * `DATE_FORMAT`. If a string, then `translateFormat` will be used as the format
  * string to determine the output; see the `formatTime` function for details.
+ * @param {function} input.colorizer A user-supplied formatter for altering output
```
jsumners, line 428:
```suggestion
 * @param {function} [input.prettifier] A user-supplied formatter for altering output.
```

lib/utils.js
pearofducks: Added! Thanks so much for your help getting this PR put in order :)

mcollina (approved):
lgtm

Readme.md
```diff
@@ -238,6 +238,32 @@ const prettifyQuery = value => {
...
+    // on if the levelKey option is used or not.
+    // By default this will be the same numerics as the Pino default:
+    level: logLevel => `LEVEL: ${logLevel}`
+  }
+}
+```
+
+Note that prettifiers do not include any coloring, if the stock coloring on `level` is desired, it can be accomplished using the following:
+```js
+const levelColorize = require('pino-pretty/lib/colors.js')(true)
```
jsumners, line 259: I do not want to promote relying on an internal API like this. If we are going to support this, we should expose the colorizer as a property of the main export, e.g. `const  { colorizerFactory } = require('pino-pretty')`.

mcollina: Good spot, I missed it!

mcollina (changes requested):
Remove the usage of `pino-pretty/lib/colors.js` and expose it.

pearofducks:
No problem!

Readme.md
```diff
@@ -238,6 +238,33 @@ const prettifyQuery = value => {
...
+Note that prettifiers do not include any coloring, if the stock coloring on `level` is desired, it can be accomplished using the following:
+```js
+const { colorizerFactory } = require('pino-pretty')
+const levelColorize = colorizerFactory(true)
+const levelPrettifier = logLevel => `LEVEL: ${levelColorize(logLevel)}`
+//...
+{
+  customPrettifiers: { level: levelPrettifier }
+}
+```
```
jsumners, line 266:
```suggestion
Additionally, `customPrettifiers` can be used to format the time and level
outputs:
```js
{
  customPrettifiers: {
    // The argument for this function will be the same
    // string that's at the start of the log-line by default:
    time: timestamp => `🕰 ${timestamp}`,

    // The argument for the level-prettifier may vary depending
    // on if the levelKey option is used or not.
    // By default this will be the same numerics as the Pino default:
    level: logLevel => `LEVEL: ${logLevel}`
  }
}
```

Note that prettifiers do not include any coloring, if the stock coloring on
`level` is desired, it can be accomplished using the following:
```js
const { colorizerFactory } = require('pino-pretty')
const levelColorize = colorizerFactory(true)
const levelPrettifier = logLevel => `LEVEL: ${levelColorize(logLevel)}`
//...
{
  customPrettifiers: { level: levelPrettifier }
}
```
```

**Review by jsumners — COMMENTED** [review]
Last nit, and then everything looks good to me.

File: Readme.md (hunk 8 above)
Comment (jsumners) [reply Readme.md]: @pearofducks somehow the suggestion removed the example. Let's add it back:

```suggestion

Note that prettifiers do not include any coloring, if the stock coloring on
`level` is desired, it can be accomplished using the following:
```js
const { colorizerFactory } = require('pino-pretty')
const levelColorize = colorizerFactory(true)
const levelPrettifier = logLevel => `LEVEL: ${levelColorize(logLevel)}`
//...
{
  customPrettifiers: { level: levelPrettifier }
}
```
```

**Review by jsumners — APPROVED** [review]
Looks good to me. Thank you for working with the review process.

File: index.js [hunk 9]
```diff
@@ -214,4 +214,5 @@ function build (opts = {}) {
 
 module.exports = build
 module.exports.prettyFactory = prettyFactory
+module.exports.colorizerFactory = colors
```
mcollina, line 217: Could you add a test to cover this new public API?

pearofducks: Sure - I've expanded the colorizer test to ensure both exports behave the same! If there's something else you had in mind please let me know. 👍

mcollina (approved):
lgtm

PR #268, opened 2021-11-19
detect err property on log object and format it using prettifyErrorLog

### LINKED ISSUE #266: Support structure for error log object used by pino@7
pino@7 now stores a logged error object on the `err` key instead of fusing it into the top-level of the log object. However, pino-pretty only checks for the old structure (`log.type === 'Error' && log.stack`). I'd like to request that pino-pretty add support for the new structure that pino@7 uses so users don't have to implement custom handlers to achieve the error object formatting as before.

Here's an example of this scenario when using pino@7:

```js
const pino = require('pino')
const pinoPretty = require('pino-pretty')

const logger = pino({ base: undefined }, pinoPretty({ destination: 2, colorize: true }))

try {
  const obj = Object.defineProperty({}, 'foo', { value: 'bar' })
  obj.foo = 'baz'
} catch (err) {
  logger.error(err)
}
```

Here's the output we'd like to see:

```
[1637306767116] ERROR: Cannot assign to read only property 'foo' of object '#<Object>'
    TypeError: Cannot assign to read only property 'foo' of object '#<Object>'
        at Object.<anonymous> (/tmp/pino-lab/index.js:10:11)
        ...
```

Here's the output we currently see:

```
[1637306825642] ERROR: Cannot assign to read only property 'foo' of object '#<Object>'
    err: {
      "type": "TypeError",
      "message": "Cannot assign to read only property 'foo' of object '#<Object>'",
      "stack":
          TypeError: Cannot assign to read only property 'foo' of object '#<Object>'
              at Object.<anonymous> (/tmp/pino-lab/index.js:10:11)
              ...
    }
```

mojavelinux:
closes #266

### DIFF
--- a/index.js
+++ b/index.js
@@ -84,6 +84,11 @@ function prettyFactory (options) {
       if (level < minimum) return
     }
 
+    if (!(messageKey in log)) {
+      const errorKey = errorLikeObjectKeys.find((it) => 'message' in (log[it] || {}))
+      if (errorKey) log = Object.assign({ [messageKey]: log[errorKey].message }, log)
+    }
+
     const prettifiedMessage = prettifyMessage({ log, messageKey, colorizer, messageFormat, levelLabel })
 
     if (ignoreKeys) {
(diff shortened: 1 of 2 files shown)


mojavelinux:
Are we no longer pursuing this? Or is it being address by another PR? I got really busy at the end of the year and haven't had a chance to follow up. But I am willing to see this through.

PR #271, opened 2021-11-20
only output space after key in object if value does not start with eol

### LINKED ISSUE #265: Don't append space after key if prettified object begins with a newline
When the singleLine option is not enabled, the prettifyObject function always inserts a space between the key and value. However, a custom prettifier function may return a string that begins with a newline (eol character) in the case it wants to do its own formatting of the value. In this case, the prettifyObject function leaves a space at the end of the line with the key.

I'd like to request that prettifyObject only insert a space between the key and the value if the value does not start with a newline.

Here's an example of when this happens:

```js
const pino = require('pino')
const pinoPretty = require('pino-pretty')

const logger = pino({}, pinoPretty({
  destination: 2,
  customPrettifiers: {
    key: () => '\n  prettified value\n  using our own formatting',
  },
}))

logger.info({ key: 'value' }, 'hello')
```

mojavelinux:
closes #265

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -395,7 +395,7 @@ function prettifyObject ({
       if (lines === undefined) return
 
       const joinedLines = joinLinesWithIndentation({ input: lines, ident, eol })
-      result += `${ident}${keyName}: ${joinedLines}${eol}`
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

PR #272, opened 2021-11-20
append missing eol before appending prettified error object when singleLine=true

mojavelinux:
Here's how the output currently looks:

```
[1637396183072] ERROR: something bad happened    Error: something bad happened
        at Object.<anonymous> (/tmp/pino-lab/fiddle.js:16:9)
```

Here's what we expect:

```
[1637396183072] ERROR: something bad happened
    Error: something bad happened
        at Object.<anonymous> (/tmp/pino-lab/fiddle.js:16:9)
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -146,6 +146,7 @@ function prettyFactory (options) {
         ident: IDENT,
         eol: EOL
       })
+      if (singleLine) line += EOL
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

Issue #273, opened 2021-11-20
Discrepancy in TypeScript types


tearwyx:
According to the module `pino-abstract-transport` the function exported by default returns a value of type `Transform` from the module `stream`, but in the declaration file it returns type `(inputData: string | object) => string`.

Based on version 7.2.0.

mcollina:
Thanks for reporting!
Would you like to send a Pull Request to address this issue? Remember to add unit tests.

tearwyx:
@mcollina #274

SushiWaUmai:
Having the same issue with the type Prettifier
```
declare namespace PinoPretty {
  type Prettifier = (inputData: string | object) => string;
  type MessageFormatFunc = (log: LogDescriptor, messageKey: string, levelLabel: string) => string;
  type PrettyOptions = PrettyOptions_
}
```

kibertoad:
@SushiWaUmai Can you send a PR with tests?

tearwyx:
If it is the same issue, there is no need to fix it once it has been fixed in the main branch (#274) and hopefully it will get out someday.

PR #274, opened 2021-11-21
fix: update return type of function exported by default (#273)

tearwyx:


index.d.ts
```diff
@@ -121,8 +127,9 @@ interface PrettyOptions_ {
 declare namespace PinoPretty {
   type Prettifier = (inputData: string | object) => string;
   type MessageFormatFunc = (log: LogDescriptor, messageKey: string, levelLabel: string) => string;
-  type PrettyOptions = PrettyOptions_
+  type PrettyOptions = PrettyOptions_;
+  type PrettyStream = Transform & OnUnknown;
```
kibertoad, line 131: Should it be both and not either?

test/types/pino-pretty.test-d.ts
```diff
@@ -52,16 +52,11 @@ const options2: PrettyOptions = {
...
+expectType<PrettyStream>(pretty(options));
+expectType<PrettyStream>(PinoPrettyNamed(options));
+expectType<PrettyStream>(PinoPrettyDefault(options));
+expectType<PrettyStream>(PinoPrettyStar.PinoPretty(options));
+expectType<PrettyStream>(PinoPrettyStar.default(options));
+expectType<PrettyStream>(PinoPrettyCjsImport.PinoPretty(options));
+expectType<PrettyStream>(PinoPrettyCjsImport.default(options));
 expectType<any>(PinoPrettyCjs(options));
-
-expectType<string>(pretty({ foo: "bar" }));
```
kibertoad, line 66: why were these removed?

tearwyx:
```
This expression is not callable.
  Type 'Transform & OnUnknown' has no call signatures. ts(2349)
```

index.d.ts
tearwyx: If you mean intersection operands, then function `pino-pretty` returns the return value of function `pino-abstract-transport` and this return value is of this type.

kibertoad (approved):

mcollina:
cc @kibertoad

mcollina (approved):
lgtm

PR #280, opened 2021-12-08
allow sonic-boom to filter call to close on stdout

### LINKED ISSUE #279: pino-pretty breaks flushing of stdout
pino-pretty overrides the `end` function on the destination to bypass the call to `fs.close`.

```js
if (destination.fd === 1) {
  // We cannot close the output
  destination.end = function () {
    this.emit('close')
  }
}
```

In doing so, it breaks flushing. In other words, if there are messages in the buffer, they get dropped when the function emits the close event.

Since sonic-boom 2.4, the call to `fs.close` is now bypassed, yet the buffer is still flushed. Therefore, I think pino-pretty should remove this logic. Besides, the logic is incomplete anyway as it does not consider stderr.

mojavelinux:
closes #279

### DIFF
--- a/index.js
+++ b/index.js
@@ -196,13 +196,6 @@ function build (opts = {}) {
         sync: false
       })
     }
-    /* istanbul ignore else */
-    if (destination.fd === 1) {
-      // We cannot close the output
-      destination.end = function () {
-        this.emit('close')
-      }
-    }
 
     source.on('unknown', function (line) {
       destination.write(line + '\n')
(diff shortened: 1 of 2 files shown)


mojavelinux:
I split this PR into two commits. The first commit asserts the existing functionality. Then, the second commit builds on that to verify that the original assertion still works and the messages are flushed when `end()` is called on the destination.

mcollina (approved):
lgtm

PR #282, opened 2021-12-11
Allow loading common js config files by default

EchelonFour:
If your project is ESM by default, then you need to make common js files have the cjs extension to load. This just adds that to the list of files to try loading.

I have also removed the custom joycon loader, because it always tries to parse JSON now.

### DIFF
--- a/bin.js
+++ b/bin.js
@@ -19,6 +19,7 @@ const parseJSON = input => {
 const joycon = new JoyCon({
   parseJSON,
   files: [
+    'pino-pretty.config.cjs',
     'pino-pretty.config.js',
     '.pino-prettyrc',
     '.pino-prettyrc.json'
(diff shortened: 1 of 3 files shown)


bin.js
```diff
@@ -19,16 +19,13 @@ const parseJSON = input => {
 const joycon = new JoyCon({
   parseJSON,
   files: [
+    'pino-pretty.config.cjs',
```
mcollina, line 22: I would add also `mjs`.

jsumners:
> I have also removed the custom joycon loader, because it always tries to parse JSON now.

For others wondering about this, `joycon@3.1.1` introduced a major change that always parses the read in file as JSON unless a specific loader is used.

bin.js
EchelonFour: mjs files can only be loaded dynamicly using `import()`, which is asynchronous. Joycon is synchronous and uses `require()`. So that can't really work. At least not simply.

mcollina (approved):
lgtm

PR #286, opened 2021-12-16
Fix cloning errors

### LINKED ISSUE #260: customPrettifiers error doesn't work with ignore option
Running this example:

```
const pino = require('pino')

const logger = pino({
	'prettyPrint': {
		'ignore': 'random',
		'customPrettifiers': {
			'error': (err) => {
				console.log('given', err); // err is {}
				
				return err.stack ?? err.message ?? err; // this will be {}
			}
		}
	},
});

logger.error({'error': new Error('message')});
```

Gives this error:
```
const lines = input.split(/\r?\n/)
                      ^

TypeError: input.split is not a function
at joinLinesWithIndentation (%PROJECT%/pino-pretty/lib/utils.js:144:23)
at prettifyError (%PROJECT%/pino-pretty/lib/utils.js:459:23)
at %PROJECT%/pino-pretty/lib/utils.js:409:15
at Array.forEach (<anonymous>)
    at prettifyObject (%PROJECT%/pino-pretty/lib/utils.js:401:26)
    at pretty (%PROJECT%/pino-pretty/index.js:152:32)
    at Object.write (%PROJECT%/pino/lib/tools.js:320:25)
    at Pino.write (%PROJECT%/pino/lib/proto.js:190:10)
    at Pino.LOG [as error] (%PROJECT%/pino/lib/tools.js:58:21)
    at Object.<anonymous> (%PROJECT%/example.js:16:8)
```

The problem is that when the filterLog method clones the given log params the value of log.error will be an empty object

Allain55:
Fixes #260

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -1,6 +1,6 @@
 'use strict'
 
-const clone = require('rfdc')({ circles: true })
+const fastCopy = require('fast-copy')
 const dateformat = require('dateformat')
 const stringifySafe = require('fast-safe-stringify')
 const defaultColorizer = require('./colors')()
@@ -564,12 +564,12 @@ function deleteLogProperty (log, property) {
  * Filter a log object by removing any ignored keys.
  *
  * @param {object} log The log object to be modified.
- * @param {string} ignoreKeys An array of strings identifying the properties to be removed.
+ * @param {Set<string> | Array<string>} ignoreKeys An array of strings identifying the properties to be removed.
  *
  * @returns {object} A new `log` object instance that does not include the ignored keys.
  */
 function filterLog (log, ignoreKeys) {
-  const logCopy = clone(log)
+  const logCopy = fastCopy(log)
   ignoreKeys.forEach((ignoreKey) => {
     deleteLogProperty(logCopy, ignoreKey)
   })
--- a/package.json
+++ b/package.json
@@ -35,6 +35,7 @@
     "args": "^5.0.1",
     "colorette": "^2.0.7",
     "dateformat": "^4.6.3",
+    "fast-copy": "^2.1.1",
     "fast-safe-stringify": "^2.0.7",
     "joycon": "^3.1.1",
     "pino-abstract-transport": "^0.5.0",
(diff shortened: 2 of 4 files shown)


mcollina (approved):
lgtm

lib/utils.js
```diff
@@ -1,6 +1,6 @@
 'use strict'
 
-const clone = require('rfdc')({ circles: true })
```
jsumners, line 3: Are there any other methods for resolving the issue at hand? This replacement module has not been updated in quite some time, has a long open dependabot PR, and https://github.com/planttheidea/fast-copy/issues/39#issuecomment-766400371

lib/utils.js
```diff
@@ -564,12 +564,12 @@ function deleteLogProperty (log, property) {
  * Filter a log object by removing any ignored keys.
  *
  * @param {object} log The log object to be modified.
- * @param {string} ignoreKeys An array of strings identifying the properties to be removed.
+ * @param {Set<string> | Array<string>} ignoreKeys An array of strings identifying the properties to be removed.
```
jsumners, line 567:
```suggestion
 * @param {string[]} ignoreKeys An array of strings identifying the properties to be removed.
```

jsumners (changes requested):

lib/utils.js
Allain55: I checked every library with deep cloning and a minimal user base but I couldn't find any other library that can clone errors and would pass every test

jsumners: Is there any other way, within this code base, outside of a cloning module?

kibertoad: @davidmarkclements ping?

mcollina (approved):
lgtm

lib/utils.js
mcollina: I think we can ship this

mcollina: If I get a fix on rdfc, I'd be very happy to.

kibertoad: which fix are you talking about?

kibertoad (approved):

lib/utils.js
mcollina: add support to rdfc to clone errors.

PR #297, opened 2022-01-19
Add support for sync: true as option

mcollina:
Fixes: https://github.com/pinojs/pino-pretty/issues/296

PR #298, opened 2022-01-21
Add type definition for the 'sync' option

OnkelTem:
Adding types for #297 

cc @mcollina  :)

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -102,6 +102,11 @@ interface PrettyOptions_ {
    */
   ignore?: string;
   /**
+   * Makes messaging synchronous.
+   * @default false
+   */
+  sync?: boolean;
+  /**
    * Provides the ability to add a custom prettify function for specific log properties.
    * `customPrettifiers` is an object, where keys are log properties that will be prettified
    * and value is the prettify function itself.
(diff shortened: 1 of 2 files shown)


mcollina (changes requested):
Can you please add a unit test? We use tsd.

OnkelTem:
Matteo, following `test/types/pino-pretty.test-d.ts` style, it would be just:

```patch
diff --git a/test/types/pino-pretty.test-d.ts b/test/types/pino-pretty.test-d.ts
index 72bdd8c..90a3405 100644
--- a/test/types/pino-pretty.test-d.ts
+++ b/test/types/pino-pretty.test-d.ts
@@ -27,7 +27,8 @@ const options: PinoPretty.PrettyOptions = {
     key: (value) => {
       return value.toString().toUpperCase();
     }
-  }
+  },
+  sync: false,
 };
 
 const options2: PrettyOptions = {
@@ -49,7 +50,8 @@ const options2: PrettyOptions = {
     key: (value) => {
       return value.toString().toUpperCase();
     }
-  }
+  },
+  sync: false,
 };
 
 expectType<PrettyStream>(pretty(options));
```

Did I get it right and that's what you meant?

mcollina:
Yes, exactly

OnkelTem:
Hi! Have I done it already? Or I missed something? Not quite sure what I'm looking at: 

image

mcollina (approved):
lgtm

Issue #305, opened 2022-02-24
`stdout is not a tty` on git bash on Windows


allevo:
Hi,
I tried to run the below command in git bash under Windows 11 and an error is shown as described below.

```
$ node http-fastify-complete.mjs | pino-pretty
stdout is not a tty
```

In fact the git bash is not a tty as desccribed below.
```
$ node -e 'console.log(tty.isatty())'
false
```

Is there any change to use pino-pretty in git bash?

Thanks

mcollina:
Have you tracked down where that error come from? I don't think it comes from pino-pretty.

allevo:
you are right.
```
$ echo 'foo' | pino-pretty
foo
```
works fine.

so, probable it is a problem of `node` itself?

allevo:
I don't think so
```
$ node -e 'console.log(666)' | pino-pretty
stdout is not a tty

```

mcollina:
Does `node -e 'console.log(666)'` work?

allevo:
yes
```
$ node -e 'console.log(666)'
666

```

mcollina:
Unfortunately I do not have a windows machine to debug this :(. If it's on pino or pino-pretty, could you send a PR?

allevo:
seems to be a worse scenario
```
$ echo 'foo' | node -e 'process.stdin.pipe(process.stout)'
stdin is not a tty

```

Any suggestion? Just where / how to proceed here: I really don't know if it is a bad stuff on my shell, nodejs problems or others....

jsumners:
I do not believe Bash as delivered by git for Windows is a true posix implementation. I _think_ it is based on Cygwin. You should try LFS for Windows.

allevo:
> I do not believe Bash as delivered by git for Windows is a true posix implementation. I _think_ it is based on Cygwin. You should try LFS for Windows.

I don't understand, sorry. Do you mean LargeFileSystem for Windows? Do you have some links?

jsumners:
This https://docs.microsoft.com/en-us/windows/wsl/about

Windows is a video game OS for me 🤷‍♂️

allevo:
Really thanks for your answers.
Anyway, with PowerShell all the above commands work fine.
So probably, this issue happens only with git bash.

Can I send a PR for updating the documentation?
Probably it is better to have a little section for advising people to use *right* shells (but I don't know what in this case means *right*....)

jsumners:
[...]

PR #306, opened 2022-02-24
doc: Shell limitation

allevo:
As discussed in https://github.com/pinojs/pino-pretty/issues/305

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -337,10 +337,13 @@ This option can also be defined as a `function` with this prototype:
 ```
 
 ## Limitations
-Because `pino-pretty` uses the stdout redirection, in some cases the command terminates with error due the shell limitations.
 
-For example, currently, mingw64 shell (installed by git bash on Windows) is affected by this issue terminating the process with `stdout is not a tty` error message
+Because `pino-pretty` uses stdout redirection, in some cases the command may
+terminate with an error due to shell limitations.
 
+For example, currently, mingw64 based shells (e.g. Bash as supplied by [git for
+Windows](https://gitforwindows.org)) are affected and terminate the process with
+a `stdout is not a tty` error message.
 
 <a id="license"></a>
 ## License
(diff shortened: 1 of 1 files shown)


mcollina (approved):
lgtm

I still think this is fixable, I just don't know where the problem is and I have no way to debug it. I think a session with a debugger could fix it.

allevo:
I really doesn't have enough experience for deep debug this issue. But I'm open to share my screen 😄 or send me some links

Readme.md
```diff
@@ -336,7 +336,13 @@ This option can also be defined as a `function` with this prototype:
...
 ```
 
-<a id="license"><a>
+## Limitations
+Because `pino-pretty` uses the stdout redirection, in some cases the command terminates with error due the shell limitations.
+
+For example, currently, mingw64 shell (installed by git bash on Windows) is affected by this issue terminating the process with `stdout is not a tty` error message
+
+
+<a id="license"></a>
```
jsumners, line 345:
```suggestion
## Limitations

Because `pino-pretty` uses stdout redirection, in some cases the command may
terminate with an error due to shell limitations.

For example, currently, mingw64 based shells (e.g. Bash as supplied by [git for
Windows](https://gitforwindows.org)) are affected and terminate the process with
a `stdout is not a tty` error message.

<a id="license"></a>
```

Is the added link correct?

jsumners (changes requested):

Readme.md
allevo: The anchor "license" exists before this commit. I just change from `<a>` to `</a>` for closing the html tag correctly.

jsumners: No. I'm asking if the link to the git package I added in the suggestion is correct.

allevo: yes it is correct

jsumners (approved):

mcollina:
How about adding a sentence saying that a PR from a Windows expert would be useful?

allevo:
Updated as request

PR #317, opened 2022-03-19
Add new option to fallback custom levels and colors to default values

bitDaft:
Currently the customLevels and customColors options completely override the default Levels.

This PR changes the default behaviour of either using customLevels or default levels, and instead makes it so that it will now override the default levels with new Levels.

The reason for this change is that there is a use case where we may only want to add a few extra custom levels and need to use the existing levels too. 

refering to [comment](https://github.com/pinojs/pino-pretty/issues/62#issuecomment-526982856) in issue #62


mcollina (changes requested):
I don't think we can accept this breaking change. Could you add new options instead?

bitDaft:
Refactored to maintain backward compatibility.

Now uses a new option to switch whether to fallback level and color option to default values

The option name is `useOnlyCustomProps`, it's default value is `true`; to get the fallback, `false` can be passed as value.

The default is kept as `true`, to keep inline with how pino uses the `useOnlyCustomLevels` option which needs to be true to override the levels. Maybe this option can be set to default `false` in the next major version to follow along with pino convention of overriding and adding custom levels

mcollina (changes requested):
Good work! Could you add the new option to the README and update the PR title?

bitDaft:
updated

mcollina (approved):
lgtm

bitDaft:
@mcollina can this PR be merged?

PR #327, opened 2022-04-07
Fixes type declarations to allow for no options, closes #326

### LINKED ISSUE #326: pretty() options required in types
Hey folks, I'm following the readme example

```typescript
const pino = require('pino')
const pretty = require('pino-pretty')
const logger = pino(pretty())
```

but I'm hitting the following type-check issue

```
src/index.ts:14:14 - error TS2554: Expected 1 arguments, but got 0.

14 app.use(pino(pretty()));
                ~~~~~~~~

  node_modules/pino-pretty/index.d.ts:16:29
    16 declare function PinoPretty(options: PrettyOptions_): PinoPretty.PrettyStream;
                                   ~~~~~~~~~~~~~~~~~~~~~~~
    An argument for 'options' was not provided.


Found 1 error in src/index.ts:14
```

which looks like from here where the options are required

https://github.com/pinojs/pino-pretty/blob/077216cb7287ff9f52980de8c61c44085a7112b5/index.d.ts#L16

I think the options argument should be optionally typed, otherwise typescript complains. Am I missing something, or is this an oversight?

As a workaround for now, this works

```typescript
app.use(pino(pretty({})));
```

daniel-j-h:
For #326. cc @mcollina

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -13,7 +13,7 @@ import { DestinationStream } from 'pino';
 
 type LogDescriptor = Record<string, unknown>;
 
-declare function PinoPretty(options: PrettyOptions_): PinoPretty.PrettyStream;
+declare function PinoPretty(options?: PrettyOptions_): PinoPretty.PrettyStream;
 
 interface PrettyOptions_ {
   /**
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

PR #340, opened 2022-06-03
add minimumLevel to typings

### LINKED ISSUE #339: Typings missing `minimumLevel`
`minimumLevel` is a valid option here:
https://github.com/pinojs/pino-pretty/blob/e793e646a0b539b3a9aa8b5ab43e64a8d86b05d6/index.js#L57
But it's not available in the typings:

Which I assume is because it isn't part of the default options set, so when the typings were written it was missed:
https://github.com/pinojs/pino-pretty/blob/e793e646a0b539b3a9aa8b5ab43e64a8d86b05d6/index.js#L30-L48

(psst... what is the default minimum level?)

LewisTehMinerz:
Resolves #339

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -9,7 +9,7 @@
 import { Transform } from 'stream';
 import { OnUnknown } from 'pino-abstract-transport';
 // @ts-ignore fall back to any if pino is not available, i.e. when running pino tests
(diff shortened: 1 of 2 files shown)


kibertoad (approved):

kibertoad:
thank you!

Issue #336, opened 2022-06-01
Drop args in favor of minimist


mcollina:
Unfortunately `args` broke our tests in v5.0.1 (ref https://github.com/pinojs/pino-pretty/pull/337)

I would recommend we drop `args` in favor of using plain old minimist.

Uzlopak:
@mcollina 

I guess that this can be closed.

PR #350, opened 2022-06-13
Dropping args and replacing by minimist

marcelfranca:
Fix issue https://github.com/pinojs/pino-pretty/issues/336


marcelfranca:
I might understood wrong but I was told to create a DRAFT as soon as I started to work to show that someone is working on it.

simoneb:
@jsumners we are used to opening Draft PRs early on, no need to request changes on a Draft PR anyway since it can't be merged.

@marcelfranca you mentioned you're having troubles migrating some options from args to minimist, shout if you need any input so the people contributing to this repo can help out.

simoneb (approved):

jsumners:
In the future, please start work from a feature branch. See https://jrfom.com/posts/2017/03/08/a-primer-on-contributing-to-projects-with-git/#create-a-feature-branch

jsumners (dismissed):
This PR results in a loss of CLI help. Prior to this PR, one could `pino-pretty --help` and get descriptions of each allowed option along with some helpful examples. With this PR, `pino-pretty --help` just hangs waiting for input.

If `minimist` does not support this feature, and newer versions of `args` have not resolved the problem prompting issue #336, then I recommend looking at `sade` instead -- https://github.com/lukeed/sade#single-command-mode

marcelfranca:
Yes, `minimist` does support, I'll add it.

Also, I created a PR for `args` that fixes the problem we had with it, but no updates so far.

Here is the link for the PR:
https://github.com/leo/args/pull/166

jsumners:
> Also, I created a PR for `args` that fixes the problem we had with it, but no updates so far.
> 
> Here is the link for the PR: [leo/args#166](https://github.com/leo/args/pull/166)

You should add a test to that PR that fails prior to your fix.

simoneb (approved):

jsumners (commented):
I don't like moving the help text out to a text file:

1. It _will_ get out of sync
2. The examples are no longer contextual, they are only available in the root `--help`

But I'm not going to block this.

PR #366, opened 2022-08-10
Line simplification for development

mcollina:
We say that `pino-pretty` is a _development_ tool, however it's very verbose in its default settings to use for local development. In this PR I simplified it a bit:

1. Use the local timezone and just print the time. If I'm developing code I want to reason on my own timezone and I'm not interested in what day it is.
2. remove the hostname as it's unlikely to change given that's my development machine.

This was inspired by https://github.com/fastify/fastify-cli/issues/536

### DIFF
--- a/test/basic.test.js
+++ b/test/basic.test.js
@@ -1,5 +1,7 @@
 'use strict'
 
+process.env.TZ = 'UTC'
+
 const { Writable } = require('readable-stream')
 const os = require('os')
 const test = require('tap').test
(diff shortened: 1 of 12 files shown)


test/basic.test.js
```diff
@@ -25,9 +27,11 @@ function prettyFactory (opts) {
   return _prettyFactory(opts)
 }
 
+// Date.prototype.getTimezoneOffset = function () { return 0 }
```
jsumners, line 30:
```suggestion
```

jsumners (approved):
Looks good to me.

PR #371, opened 2022-08-19
chore: update example image

Fdawgs:
The original example image is over 4 years old and no longer accurately reflects what pino-pretty does, i.e. v9.0.0 removed the date from the output.


mcollina (approved):
lgtm

PR #373, opened 2022-08-21
A new option for users to choose log keys to include

### LINKED ISSUE #363: [Feature] Add option which does the opposite of `ignore`
Our use case of pino is to include contextual info under 1 log key (like `data`). During development, we're only really interested in this key, along with the top log line output (timestamp, message, etc).

The options for pino-pretty don't allow us to ignore all keys except for `data`, unless we ignore keys exhaustively. This means we have to understand all possible keys in the object, and have to keep updating this list as we find more keys to ignore. This isn't really sustainable.

## Proposal

Add an option `only`, or `select`, which will select the keys to output, and will override `ignore` if they are both given.

TommyDew42:
Closes #363.
* The new option is named `include`.
* `--ignore` will be ignored if `--include` is passed in.
* CLI long flag is `--include` and short flag `--I`.

### DIFF
--- a/help/help.txt
+++ b/help/help.txt
@@ -15,6 +15,7 @@
     -h, --help                  Output usage information
     -H, --hideObject            Hide objects from output (but not error object)
     -i, --ignore                Ignore one or several keys: (`-i time,hostname`)
+    -I, --include               The opposite of `--ignore`, only include one or several keys: (`-I level,time`)
     -l, --levelFirst            Display the log level as the first output field
     -L, --levelKey [value]      Detect the log level under the specified key (defaults to "level")
     -b, --levelLabel [value]    Output the log level using the specified label (defaults to "levelLabel")
@@ -57,6 +58,9 @@
 
     - Prettify logs but don't print pid and hostname
     $ cat log | pino-pretty -i pid,hostname
+    
+    - Prettify logs but only print time and level
+    $ cat log | pino-pretty -i time,level
 
     - Loads options from a config file
     $ cat log | pino-pretty --config=/path/to/config.json
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -571,15 +571,27 @@ function deleteLogProperty (log, property) {
 }
 
 /**
- * Filter a log object by removing any ignored keys.
+ * Filter a log object by removing or including keys accordingly.
+ * When `includeKeys` is passed, `ignoredKeys` will be ignored.
  *
  * @param {object} log The log object to be modified.
  * @param {Set<string> | Array<string>} ignoreKeys An array of strings identifying the properties to be removed.
+ * @param {Set<string> | Array<string>} includeKeys An array of strings identifying the properties to be included.
  *
  * @returns {object} A new `log` object instance that does not include the ignored keys.
  */
-function filterLog (log, ignoreKeys) {
+function filterLog (log, ignoreKeys, includeKeys) {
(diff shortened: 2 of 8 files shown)


mcollina (approved):
lgtm

lib/utils.js
```diff
@@ -571,15 +571,27 @@ function deleteLogProperty (log, property) {
...
+ * When `includeKeys` is passed, `ignoredKeys` will be ignored.
  *
  * @param {object} log The log object to be modified.
  * @param {Set<string> | Array<string>} ignoreKeys An array of strings identifying the properties to be removed.
+ * @param {Set<string> | Array<string>} includeKeys An array of strings identifying the properties to be included.
  *
  * @returns {object} A new `log` object instance that does not include the ignored keys.
  */
-function filterLog (log, ignoreKeys) {
+function filterLog (log, ignoreKeys, includeKeys) {
```
jsumners, line 583: I'd rather change this signature. Clearly we are going to keep adding to it.

```suggestion
function filterLog ({ log, ignoreKeys = [], includeKeys = [] } = {}) {
```

jsumners (changes requested):
https://github.com/pinojs/pino-pretty/pull/350#pullrequestreview-1015029044

mcollina (approved):
lgtm

lib/utils.js
TommyDew42: Also updated the `help.txt` in 2bf5c72.

lib/utils.js
```diff
@@ -571,15 +571,30 @@ function deleteLogProperty (log, property) {
...
+ * @param {Set<string> | Array<string> | undefined} ignoreKeys An array of strings identifying the properties to be removed.
+ * @param {Set<string> | Array<string> | undefined} includeKeys An array of strings identifying the properties to be included.
  *
- * @returns {object} A new `log` object instance that does not include the ignored keys.
+ * @returns {object} A new `log` object instance that
+ *  either only includes the keys in ignoreKeys
+ *  or does not include those in ignoredKeys.
  */
-function filterLog (log, ignoreKeys) {
+function filterLog ({ log, ignoreKeys, includeKeys }) {
```
jsumners, line 586: Ugh, GitHub won't let me provide a suggestion here due to "deleted lines". We need to tweak the docs to match the new signature:

```js
/**
 * Stuff
 *
 * @param {object} input
 * @param {object} input.log The log object to be modified.
 * @param {...} [input.ignoreKeys] 
 * etc...
 */

Comment (TommyDew42) [reply lib/utils.js]: Ah so sorry i missed that!

**Review by jsumners — APPROVED** [review]

**Comment by glensc** [comment]
The short flag is `-I` not `--I` (incorrect in pr body)

File: help/help.txt [hunk 3]
```diff
@@ -57,6 +58,9 @@
 
     - Prettify logs but don't print pid and hostname
     $ cat log | pino-pretty -i pid,hostname
+    
+    - Prettify logs but only print time and level
+    $ cat log | pino-pretty -i time,level
```
glensc, line 63: i'm sure the example should be here `-I` not `-i`

TommyDew42: Thanks for catching that! Would you wanna send a PR for that? If not, i can fix that.

glensc: i don't wanna send pr :)

TommyDew42: PR up in #375

PR #375, opened 2022-08-30
Fix the include flag example in help.txt

TommyDew42:
Fix the wrong example of the include flag in help.txt.

Thanks to @glensc spotting the issue in a previous [PR](https://github.com/pinojs/pino-pretty/pull/373/files#r958686611).


mcollina (approved):
lgtm

PR #379, opened 2022-10-03
upgrade: fast-copy to v3

ilteoood:
Upgrade fast-copy to v3.

This PR will also unlock https://github.com/pinojs/pino-webpack-plugin/pull/54

cc: @simoneb

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -1,6 +1,6 @@
 'use strict'
 
-const fastCopy = require('fast-copy')
+const fastCopy = require('fast-copy').default
 const dateformat = require('dateformat')
 const SonicBoom = require('sonic-boom')
 const stringifySafe = require('fast-safe-stringify')
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
   "dependencies": {
     "colorette": "^2.0.7",
     "dateformat": "^4.6.3",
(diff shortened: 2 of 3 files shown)


simoneb (approved):
fast-copy nonetheless should export properly so that we don't have to import as .default

jsumners (commented):
What does this fix? It seems like it will be a semver major for `pino-pretty`?

jsumners:
I agree with @simoneb. The module should export such that `const foo = require('foo')` works. Then it's a semver range update only.

lib/utils.js
```diff
@@ -1,6 +1,6 @@
 'use strict'
 
-const fastCopy = require('fast-copy')
+const { createCopier } = require('fast-copy')
```
jsumners, line 3: How is this better than the `.default` version? We are still changing code to satisfy a dependency.

simoneb: fast-copy released a semver major with this change. we need the new version to unblock https://github.com/pinojs/pino-webpack-plugin/pull/54, so this is the only solution we have, which looks fairly reasonable to me

jsumners (approved):

simoneb:
@ilteoood can you do the follow up once a new version is released please?

Issue #399, opened 2023-02-06
Change the way of passing customLevels in code level


ademarsj:
Due to `pino` handle customLevels as object often people will try to use like that in `pino-pretty`, so receiving a object instead of a string when using in code level as a transporter in `pino` would improve the understanding and use of the lib.

To work like that, instead of transforming the object to string and keeping two instances who had the same value but in different types:

```
const levels = {
  http: 10,
  info: 30,
  debug: 35,
  warn: 40,
  normal_error: 50,
  fatal_error: 60,
}

const log = pino(
  {
    customLevels: levels,
    useOnlyCustomLevels: true,
    level: 'http'
  },
  pretty({
  colorize: true,
  levelFirst: true,
  customLevels: levels,
  })
);
```

ademarsj:
I would like to work on that, so i will send a PR as soon as i can.

ademarsj:
I would like to work on that, so i will send a PR as soon as i can. (Sorry for closing the PR, i missed the click...)

PR #400, opened 2023-02-07
feat: adding possibility to pass an object as parameter to customLevels

ademarsj:
Hello, i made the changes to make possible pass in code as object or string the customLevels prop, it's still compatible with the CLI passing as string parameter.  I already wrote the tests too and keeps on the 100% coverage and asserted.

Is relative to #399 .

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -29,7 +29,9 @@ module.exports = {
   prettifyObject,
   prettifyTime,
   buildSafeSonicBoom,
-  filterLog
+  filterLog,
+  handleCustomlevelsOpts,
+  handleCustomlevelNamesOpts
 }
 
 module.exports.internals = {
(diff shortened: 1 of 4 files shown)


lib/utils.js
```diff
@@ -669,3 +671,64 @@ function autoEnd (stream, eventName) {
...
+  if (typeof cLevels === 'string') {
+    return cLevels
+      .split(',')
+      .reduce((agg, value, idx) => {
+        const [levelName, levelIdx = idx] = value.split(':')
+        agg[levelIdx] = levelName.toUpperCase()
+        return agg
+      },
+      { default: 'USERLVL' })
+  } else if (typeof cLevels === 'object') {
```
jsumners, line 694: An instance of `[]` would pass this check. The correct check is:
```suggestion
  } else if (Object.prototype.toString.call(cLevels) === '[object Object]') {
```

lib/utils.js
```diff
@@ -669,3 +671,64 @@ function autoEnd (stream, eventName) {
...
 }
+
+/**
+ * Returns a object which will be used to handle custom levels
+ *
+ * @param {string|object} cLevels Custom Levels can be object or string
+ *
+ * @returns {object} A object like level: level_name
+ */
+function handleCustomlevelsOpts (cLevels) {
```
jsumners, line 682: Thank you for providing jsdoc for the new function. However, the documentation should provide some real value over what the function signature already provides:

```suggestion
/**
 * Parse a CSV string or options object that specifies
 * configuration for custom levels.
 *
 * @param {string|object} cLevels An object mapping standard level
 * names to new level names, e.g. `{ info: 'information') }`, or a
 * CSV string in the format `level:new_name`, e.g.
 * `info:information,debug:devel`.
 *
 * @returns {object} A object mapping known level names to customized
 * level names.
 */
function handleCustomlevelsOpts (cLevels) {
```

lib/utils.js
```diff
@@ -669,3 +671,64 @@ function autoEnd (stream, eventName) {
...
+  }
+}
+
+/**
+ * Returns a object which will be used to handle custom levels
+ *
+ * @param {string|object} cLevels Custom Levels can be object or string
+ *
+ * @returns {object} A object like level_name: level
+ */
```
jsumners, line 712: Please adjust this docblock in the same manner as my suggestion on the other one.

jsumners (changes requested):
Looking good. I have some requested changes, and the documentation around the feature should also be updated.

lib/utils.js
ademarsj: True, i never did like that, i will change both checks, thank you.

lib/utils.js
ademarsj: Will be done.

lib/utils.js
ademarsj: Ok

ademarsj:
I appreciate the request and comments, i fixed them, but i got some doubts/points:
- I believe that you just give me an example of the method docs, so i changed a little to something more strict to the function returns
- I noticed that in `handleCustomlevelNamesOpts` method when passing an object as parameter the return is practically the same object that is passed, but i didn't remove the logic because of the lower case transformation, do you think it's appropriate to leave it that way ? Or just return the object as it came (expecting that the user will always pass an object with properties names already in lower case), and pass to lower case only in string mode ?
- I added the docs to the PrettyOptions with examples, I hope I made right

jsumners (changes requested):

jsumners:
> * I noticed that in `handleCustomlevelNamesOpts` method when passing an object as parameter the return is practically the same object that is passed, but i didn't remove the logic because of the lower case transformation, do you think it's appropriate to leave it that way ? Or just return the object as it came (expecting that the user will always pass an object with properties names already in lower case), and pass to lower case only in string mode ?

I would review the `git blame` for the implementation and the PRs that introduced it. The discussion therein should answer these questions.

> * I added the docs to the PrettyOptions with examples, I hope I made right

I don't understand. What is `PrettyOptions`?

ademarsj:
> I would review the `git blame` for the implementation and the PRs that introduced it. The discussion therein should answer these questions.

I did it, in this PR,  as it was just passed as a string it always will transform to lower case to make the attributes names to the object returned, now it will receive an object, if the object received it's already with lower case in his attributes names the for loop is useless, once it will return an object same as received.

> I don't understand. What is `PrettyOptions`?
The name of the Interface in index.d.ts which describes the pino-pretty options.

I will fix the English errors that you appointed soon.

ademarsj:
Done, sorry for the bad english.

jsumners (approved):
Looks good to me.

jsumners:
> Done, sorry for the bad english.

No worries. That's what reviews are for.

PR #403, opened 2023-02-20
Add a colorizeObjects option to disable object colorization

Avaq:
This allows log parsers that can detect JSON at the end of a log line to do so without being thrown off by ANSI colors.

Specifically, we're using Pino + Pino Pretty with Papertrail. Papertrail has [this feature](https://www.papertrail.com/help/json-search/) where they parse JSON at the end of a line. But their JSON parser fails to do so when that JSON is surrounded by ANSI codes.

I could disable colorization altogether, but Papertrail does nicely present ANSI colored logs in their web-UI, so I don't want to do that.

I've sent the Papertrail devs a message, because I think it'd be nicest if it were solved on their side, but as a plan B, and for the time being, this is the next best solution.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -57,6 +57,7 @@ node app.js | pino-pretty
 ### CLI Arguments
 
 - `--colorize` (`-c`): Adds terminal color escape sequences to the output.
+- `--colorizeObjects` (`-C`): Allows suppressing colorization of objects when set to `false`. In combination with `--singleLine`, this ensures that the end of each line is parsable JSON.
 - `--crlf` (`-f`): Appends carriage return and line feed, instead of just a line
   feed, to the formatted log line.
 - `--errorProps` (`-e`): When formatting an error object, display this list
@@ -235,6 +236,7 @@ The options accepted have keys corresponding to the options described in [CLI Ar
 ```js
 {
   colorize: colorette.isColorSupported, // --colorize
+  colorizeObjects: true, //--colorizeObjects
   crlf: false, // --crlf
   errorLikeObjectKeys: ['err', 'error'], // --errorLikeObjectKeys
   errorProps: '', // --errorProps
--- a/bin.js
+++ b/bin.js
@@ -40,6 +40,7 @@ const DEFAULT_VALUE = '\0default'
 let opts = minimist(process.argv, {
   alias: {
     colorize: 'c',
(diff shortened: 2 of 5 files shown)


jsumners:
I'm confused. Why are you sending the output of `pino-pretty` to a log collector?

Avaq:
We don't need to merge this. I can just use the fork for now and see what comes out of the Papertrail support ticket. Also, I understand that this change doesn't really align with the JSON logger philosophy: That disabling the colors on objects is really just muddying the separation of data and presentation (we're using some part of the presentation as data).

Avaq:
For reference, here's what the 3 situations look like in Papertrail:

- **Colors disabled** - JSON gets parsed, but no colors
  latest-screenshot
- **Colors enabled** - JSON not parsed, but colors
  latest-screenshot
- **Colors enabled, but disabled for objects** - JSON gets parsed, and colors
  latest-screenshot

And when sending it unformatted JSON logs, it'll just show the whole JSON object printed as shown in the screenshots, without any way for the user to control how it's formatted.

I don't know if this issue is worth solving here. I just opened the PR because I had the code ready, and it might be useful to some folks.

mcollina (approved):
lgtm

mcollina:
(Note that you shouldn't really do what you are doing for the optimal production setup)

mcollina:
@jsumners PTAL

jsumners (changes requested):
Full documentation is missing. I would imagine it should go between the current lines 59 and 60 of the readme.

Avaq:
@jsumners I [pushed](https://github.com/pinojs/pino-pretty/compare/8554b4a82bbb9dddd0b22f8e6ad16b06d4af1164..5ef89c78532d04d635d83a070ad5a355bf31e9ba) a version with updated docs. Sorry I missed this earlier.

Avaq:
I also heard back from Papertrail. They say it'll take a lot to fix on their end, and they proposed that a change to pino-pretty might be the easiest.

jsumners:
> I also heard back from Papertrail. They say it'll take a lot to fix on their end, and they proposed that a change to pino-pretty might be the easiest.

Of course they will put the onus on the open source community. They have zero incentive to improve their product.

jsumners (approved):

PR #404, opened 2023-02-21
fix: node --disable-proto=throw compatibility

castarco:
Bumps required (patch) version for `fast-copy` to ensure compatibility with NodeJS's `--disable-proto=throw` option (useful to mitigate prototype pollution attacks).

Reference: https://github.com/planttheidea/fast-copy/releases/tag/v3.0.1


mcollina:
There is no need for this change, just update your dependencies.

PR #407, opened 2023-03-04
Support nested levelKey

nikrabaev:
This PR allows specifying the nested `levelKey`. Such keys will be treated just like the `ignore` nested keys.
There are two reasons for this change:
1. Why would the `ignore` option support nested keys and `levelKey` don't?
2. I personally need it because in my project I use the custom format of logs, where the `level` is placed inside another `tags` object

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -68,8 +68,8 @@ node app.js | pino-pretty
   error like objects. Default: `err,error`.
 - `--messageKey` (`-m`): Define the key that contains the main log message.
   Default: `msg`.
-- `--levelKey` (`--levelKey`): Define the key that contains the level of the log, nested keys are supported with each property delimited by a dot character (`.`),
-  keys may be escaped to target property names that contains the delimiter itself:
+- `--levelKey` (`--levelKey`): Define the key that contains the level of the log. Nested keys are supported with each property delimited by a dot character (`.`).
+  Keys may be escaped to target property names that contains the delimiter itself:
   (`--levelKey tags\\.level`).
   Default: `level`.
 - `--levelLabel` (`-b`): Output the log level using the specified label.
--- a/Readme.md
+++ b/Readme.md
@@ -68,7 +68,9 @@ node app.js | pino-pretty
   error like objects. Default: `err,error`.
 - `--messageKey` (`-m`): Define the key that contains the main log message.
   Default: `msg`.
-- `--levelKey` (`--levelKey`): Define the key that contains the level of the log.
+- `--levelKey` (`--levelKey`): Define the key that contains the level of the log, nested keys are supported with each property delimited by a dot character (`.`),
+  keys may be escaped to target property names that contains the delimiter itself:
+  (`--levelKey tags\\.level`).
   Default: `level`.
 - `--levelLabel` (`-b`): Output the log level using the specified label.
   Default: `levelLabel`.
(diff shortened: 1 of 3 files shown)


mcollina (approved):
lgtm

nikrabaev:
Even though this got approved, I've just realized that it may break things for those who assumed the dot notation handing does not apply to the `levelKey` and didn't escape the dots in their code. 
I see 3 ways to avoid such issues:
1. Try to find the property using the non-split `levelKey` first. If the value is not `unknown`, return it. Otherwise, assume it's a nested key.
2. Create a new option like `nestedLevelKey` while preserving the original behavior of `levelKey`. In situations when both are used, give higher priority to the `nestedLevelKey`.
3. Mark it as a breaking change.

mcollina:
I think we might just go for a major.

@jsumners wdyt?

Readme.md
```diff
@@ -68,7 +68,9 @@ node app.js | pino-pretty
   error like objects. Default: `err,error`.
 - `--messageKey` (`-m`): Define the key that contains the main log message.
   Default: `msg`.
-- `--levelKey` (`--levelKey`): Define the key that contains the level of the log.
+- `--levelKey` (`--levelKey`): Define the key that contains the level of the log, nested keys are supported with each property delimited by a dot character (`.`),
+  keys may be escaped to target property names that contains the delimiter itself:
+  (`--levelKey tags\\.level`).
```
jsumners, line 73:
```suggestion
- `--levelKey` (`--levelKey`): Define the key that contains the level of the log. Nested keys are supported with each property delimited by a dot character (`.`).
  Keys may be escaped to target property names that contains the delimiter itself:
  (`--levelKey tags\\.level`).
```

lib/utils.js
```diff
@@ -555,6 +552,29 @@ function splitIgnoreKey (key) {
...
+ * Gets a specified property from an object if it exists.
+ * @param {object} obj The object to be searched.
+ * @param {(string|string[])} property A string or an array of strings identifying
+  * the property to be retrieved from the object.
+ * Accepts nested properties delimited by a `.`
+ * Delimiter can be escaped to preserve property names that contain the delimiter.
+ * e.g. `'prop1.prop2'` or `'prop2\.domain\.corp.prop2'`
+ * @returns {*}
+ */
+function getPropertyValue (obj, property) {
```
jsumners, line 565:
```suggestion
/**
 * Gets a specified property from an object if it exists.
 *
 * @param {object} obj The object to be searched.
 * @param {string|string[]} property A string, or an array of strings, identifying
  * the property to be retrieved from the object.
 * Accepts nested properties delimited by a `.`.
 * Delimiter can be escaped to preserve property names that contain the delimiter.
 * e.g. `'prop1.prop2'` or `'prop2\.domain\.corp.prop2'`.
 *
 * @returns {*}
 */
function getPropertyValue (obj, property) {
```

jsumners (approved):
LGTM with some doc tweaks. I'm fine with a new major.

nikrabaev:
@jsumners Applied

krazar:
I'll comment here, but tell me If it is best I create a new issue for this.

This  change may introduce an unexpected situation when using a nested level key.
Though the nested key is identified as a level, it is not removed from the object as  it is the case with "non nested levelKey" and it could be the intended behavior (at least it is mine).

Using the `exclude` options could have been a good alternative, but this is applied before loading the level property, so it does not work.

PR #434, opened 2023-06-23
fix: adds support for printing number and boolean messageKey value type

### LINKED ISSUE #433: Pino Pretty does not output numbers but only strings.
I have a very simple log file with as below `app.log`
```
{"level":30,"time":"2023-06-19T04:01:56.478Z","pid":2484,"hostname":"YMLTITD2227","msg":77}
{"level":30,"time":"2023-06-19T04:01:56.478Z","pid":2484,"hostname":"YMLTITD2227","msg":"42"}
```

Notice that first message is number 77, while the other is string 42. If I do `cat app.log | npx pino-pretty`, I get the below output
```
[12:01:56.478] INFO (2484):
[12:01:56.478] INFO (2484): 42
```

It does NOT display the numbers, but only the strings. I was hoping that pino-pretty should be able to interpret numbers as strings.

rjmohammad:
closes #433

### DIFF
--- a/test/basic.test.js
+++ b/test/basic.test.js
@@ -133,6 +133,7 @@ test('basic prettifier tests', (t) => {
     }))
     log.info('baz')
   })
+  
   t.test('can print message key value when its a number', (t) => {
     t.plan(1)
     const pretty = prettyFactory()
--- a/test/basic.test.js
+++ b/test/basic.test.js
@@ -165,6 +165,7 @@ test('basic prettifier tests', (t) => {
     }))
     log.info(true)
   })
+  
   t.test('can use different message keys', (t) => {
     t.plan(1)
     const pretty = prettyFactory({ messageKey: 'bar' })
--- a/test/basic.test.js
+++ b/test/basic.test.js
@@ -133,7 +133,7 @@ test('basic prettifier tests', (t) => {
     }))
     log.info('baz')
   })
-  
+
   t.test('can print message key value when its a number', (t) => {
     t.plan(1)
     const pretty = prettyFactory()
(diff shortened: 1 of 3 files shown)


mcollina (approved):
lgtm

test/basic.test.js
```diff
@@ -118,6 +118,51 @@ test('basic prettifier tests', (t) => {
...
+        t.equal(
+          formatted,
+          `[${formattedEpoch}] INFO (${pid}): baz\n`
+        )
+        cb()
+      }
+    }))
+    log.info('baz')
+  })
+  t.test('can print message key value when its a number', (t) => {
```
jsumners, line 136:
```suggestion
  })
  
  t.test('can print message key value when its a number', (t) => {
```

test/basic.test.js
```diff
@@ -118,6 +118,51 @@ test('basic prettifier tests', (t) => {
...
+        t.equal(
+          formatted,
+          `[${formattedEpoch}] INFO (${pid}): 42\n`
+        )
+        cb()
+      }
+    }))
+    log.info(42)
+  })
+  t.test('can print message key value when its a string', (t) => {
```
jsumners, line 151:
```suggestion
  })
  
  t.test('can print message key value when its a boolean', (t) => {
```

test/basic.test.js
```diff
@@ -118,6 +118,51 @@ test('basic prettifier tests', (t) => {
...
+        t.equal(
+          formatted,
+          `[${formattedEpoch}] INFO (${pid}): true\n`
+        )
+        cb()
+      }
+    }))
+    log.info(true)
+  })
   t.test('can use different message keys', (t) => {
```
jsumners, line 166:
```suggestion
  })
  
  t.test('can use different message keys', (t) => {
```

test/basic.test.js
```diff
@@ -118,6 +118,51 @@ test('basic prettifier tests', (t) => {
...
+      write (chunk, enc, cb) {
+        const formatted = pretty(chunk.toString())
+        t.equal(
+          formatted,
+          `[${formattedEpoch}] INFO (${pid}): true\n`
+        )
+        cb()
+      }
+    }))
+    log.info(true)
```
jsumners, line 164: I wonder what `log.info(0)` or `log.info(1)` would produce here.

jsumners (changes requested):

test/basic.test.js
rjmohammad: I've added a possible fix for 0.

What is the expectation for `log.info(null)` or `log.info(undefined`)?

jsumners: Only `null` is valid json. So I imagine we'd just print the word "null". But that can be a separate PR.

jsumners (approved):

Issue #439, opened 2023-07-19
customColors option missing from Typescript definition


Hufschmidt:
Nothing major, but the typescript interface definition `PrettyOptions_` is missing the `customColors` parameter.

mcollina:
Thanks for reporting! Would you like to send a Pull Request to address this issue? Remember to add unit tests.

Hufschmidt:
My lazy butt created a Pull Request :nerd_face:

PR #440, opened 2023-07-19
Add customColors to typescript definition for PrettyOptions

Hufschmidt:
See https://github.com/pinojs/pino-pretty/issues/439

Also contains minor cleanup of type test-file.

---

This assumes `index.d.ts` is managed manually and not autogenerated by some process as I didn't see any related script in `package.json`.

Also I've not worked with [TSD](https://github.com/SamVerschueren/tsd) so far, so I only added the missing example values for `customLevels` and `customColors` parameters to the options. 
A quick inverse check confirmed tests working as expected as they failed without adding `customColors` to the type definition. :+1:

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -189,6 +189,16 @@ interface PrettyOptions_ {
    * @example ( Object ) customLevels: { info: 10, some_level: 40 }
    */
   customLevels?: string|object;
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

Issue #441, opened 2023-07-21
Extention for --messageFormat option to support go template control structures


timlohse1104:
Hey there, 

we use the `--messageFormat` in a .pino-prettyrc file to format pino logs of a nestjs fastify service to be more readable.

The issue here would be that not every log emitted from the nestjs service is indeed based upon a http request (for example startup messages or logs emitted from a bull processor). 

Let's say we want to format logs like this without req/res objects:

`{"level":"DEBUG","time":"21.7.2023, 12:01:08","pid":32678,"hostname":"tlohse-ThinkPad-P53","context":"OcrConsumer","msg":"Job #482: Synchronous OCR analysis took 5.52 seconds."}`

and http logs like this with req/res objects:

``{"level":"INFO","time":"21.7.2023,` 12:00:56","pid":32678,"hostname":"tlohse-ThinkPad-P53","req":{"id":"req-1","method":"OPTIONS","url":"/ocr","query":{},"headers":{"host":"localhost:3005","connection":"keep-alive","accept":"*/*","access-control-request-method":"POST","access-control-request-headers":"authorization,content-type","origin":"http://localhost:4200","user-agent":"Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/114.0.0.0 Safari/537.36 Edg/114.0.1823.82","sec-fetch-mode":"cors","sec-fetch-site":"same-site","sec-fetch-dest":"empty","referer":"http://localhost:4200/","accept-encoding":"gzip, deflate, br","accept-language":"de,de-DE;q=0.9,en;q=0.8,en-GB;q=0.7,en-US;q=0.6"},"remoteAddress":"127.0.0.1","remotePort":42528},"res":{"statusCode":204,"headers":{}},"responseTime":4,"msg":"request completed"}``

with a ```--messageFormat``` like this: 

`"[Nest] ({req.id}) - {time} (UTC) | {level} | [{context}] {msg}"`

The output of the both logs would be:

1. [Nest] () - 21.7.2023, 12:01:08 (UTC) | DEBUG | [OcrConsumer] Job #482: Synchronous OCR analysis took 5.52 seconds. 
2. [Nest] (req-1) - 21.7.2023, 12:00:56 (UTC) | INFO | [] request completed

You can see what's missing here:

1. Misses the req.id because it was emitted in an async queue processor context
2. Misses the context

[...]

PR #442, opened 2023-07-23
Add support for messageFormat strings containing conditionals

timlohse1104:
See [#441](https://github.com/pinojs/pino-pretty/issues/441)
---
It's now possible to pass `--messageFormat` option containing conditional statements for log data.

Example log data:
`{level: 30, req: {id: 'foo'}}`

Example `--messageFormat` containing conditional:
`"{level} - {if req.id}({req.id}){end}{if msg} - {msg}{end}"`

Example output:
`"30 - (foo)"`

Unescaped if / end block  within `--messageFormat` will just be deleted.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -343,6 +343,14 @@ const levelPrettifier = logLevel => `LEVEL: ${levelColorize(logLevel)}`
 }
 ```
 
+In addition to this, if / end statement blocks can also be specified. Else statements and nested conditions are not supported.
+
+```js
+{
+  messageFormat: '{levelLabel} - {if pid}{pid}{end} - url:{req.url}'
+}
+```
+
 This option can also be defined as a `function` with this prototype:
 
 ```js
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -787,7 +787,7 @@ function handleCustomlevelNamesOpts (cLevels) {
 function parseMessageFormat (messageFormat, log) {
   messageFormat = messageFormat.replace(/{if (.*?)}(.*?){end}/g, function (_, key, value) {
     const propertyValue = getPropertyValue(log, key)
-    if (propertyValue) {
+    if (propertyValue && value.includes(key)) {
       return value.replace(new RegExp('{' + key + '}', 'g'), propertyValue)
     } else {
       return ''
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -263,16 +263,53 @@ function prettifyLevel ({ log, colorizer = defaultColorizer, levelKey = LEVEL_KE
  */
 function prettifyMessage ({ log, messageFormat, messageKey = MESSAGE_KEY, colorizer = defaultColorizer, levelLabel = LEVEL_LABEL, levelKey = LEVEL_KEY, customLevels, useOnlyCustomProps }) {
   if (messageFormat && typeof messageFormat === 'string') {
-    const message = String(messageFormat).replace(/{([^{}]+)}/g, function (match, p1) {
-      // return log level as string instead of int
-      let level
-      if (p1 === levelLabel && (level = getPropertyValue(log, levelKey)) !== undefined) {
-        const condition = useOnlyCustomProps ? customLevels === undefined : customLevels[level] === undefined
-        return condition ? LEVELS[level] : customLevels[level]
-      }
-      // Parse nested key access, e.g. `{keyA.subKeyB}`.
-      return getPropertyValue(log, p1) || ''
-    })
+    // console.log(
+    //   "pino-pretty | prettifyMessage | messageFormat: ",
+    //   messageFormat
+    // );
+    // console.log(
+    //   "pino-pretty | prettifyMessage | typeof messageFormat: ",
+    //   typeof messageFormat
+    // );
+    // console.log("pino-pretty | prettifyMessage | log: ", log);
+    // console.log("pino-pretty | prettifyMessage | typeof log: ", typeof log);
+
+    const parseMessageFormat = (input, log) => {
+      input = input.replace(/{if (.*?)}(.*?){end}/g, function (_, key, value) {
+        // console.log(`_: "${_}"`);
+        // console.log(`key: "${key}"`);
+        // console.log(`value: "${value}"`);
+        const propertyValue = getPropertyValue(log, key)
+        if (propertyValue) {
+          return value.replace(new RegExp('{' + key + '}', 'g'), propertyValue)
+        } else {
+          return ''
+        }
+      })
+
+      // console.log(`input: "${input}"`);
+      const replacedInput = input.replace(/\s+/g, ' ').trim()
+      // console.log(`replacedInput: "${replacedInput}"`);
+      return replacedInput
+    }
+
+    const parsedMessageFormat = parseMessageFormat(messageFormat, log)
+    // console.log("parsedMessageFormat", parsedMessageFormat);
+
+    const message = String(parsedMessageFormat).replace(
+      /{([^{}]+)}/g,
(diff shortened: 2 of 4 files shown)


test/lib/utils.internals.test.js
```diff
@@ -222,3 +222,29 @@ tap.test('#getPropertyValue', t => {
...
+
+  t.test('parseMessageFormat removes unescaped if statements', async t => {
+    const log = fastCopy(logData)
+    t.equal(internals.parseMessageFormat('{level} - {if data1.data2}{data1.data2}', log), '{level} - {data1.data2}')
+  })
+
+  t.test('parseMessageFormat removes unescaped end statements', async t => {
+    const log = fastCopy(logData)
+    t.equal(internals.parseMessageFormat('{level} - {data1.data2}{end}', log), '{level} - {data1.data2}')
+  })
```
mcollina, line 247: Could you add a few tests for not-matching if/else/end pair? What error would you get?

mcollina (changes requested):
Good work! Could you add some docs?

test/lib/utils.internals.test.js
timlohse1104: Yeah for sure. Thanks for the fast code review.

timlohse1104:
Updated documentation with `--messageFormat` condition example.
Add matching for condition and property keys.
Add further unit tests for parseMessageFormat function.

Readme.md
```diff
@@ -343,6 +343,14 @@ const levelPrettifier = logLevel => `LEVEL: ${levelColorize(logLevel)}`
 }
 ```
 
+In addition to this, if / end statement blocks can also be specified. Else statements and nested conditions are not supported.
+
+```js
+{
+  messageFormat: '{levelLabel} - {if pid}{pid}{end} - url:{req.url}'
```
mcollina, line 350:
```suggestion
  messageFormat: '{levelLabel} - {if pid}{pid} - {end}url:{req.url}'
```

timlohse1104: Agreed. This would be a better example to showcase the condition option.

lib/utils.js
```diff
@@ -766,3 +772,32 @@ function handleCustomlevelNamesOpts (cLevels) {
...
+    }
+  })
+
+  // Remove unescaped if blocks
+  messageFormat = messageFormat.replace(/{if (.*?)}/g, '')
+  // Remove unescaped end blocks
+  messageFormat = messageFormat.replace(/{end}/g, '')
+
+  return messageFormat.replace(/\s+/g, ' ').trim()
+}
```
jsumners, line 803: I think this would be a bit easier to read:

```suggestion
function parseMessageFormat (messageFormat, log) {
  messageFormat = messageFormat.replace(/{if (.*?)}(.*?){end}/g, replacer)

  // Remove unescaped if blocks
  messageFormat = messageFormat.replace(/{if (.*?)}/g, '')
  // Remove unescaped end blocks
  messageFormat = messageFormat.replace(/{end}/g, '')

  return messageFormat.replace(/\s+/g, ' ').trim()
  
  function replacer (_, key, value) {
    const propertyValue = getPropertyValue(log, key)
    if (propertyValue && value.includes(key)) {
      return value.replace(new RegExp('{' + key + '}', 'g'), propertyValue)
    } else {
      return ''
    }
  }
}
```

test/lib/utils.internals.test.js
```diff
@@ -222,3 +222,65 @@ tap.test('#getPropertyValue', t => {
...
+    const log = fastCopy(logData)
+    t.equal(internals.parseMessageFormat('{level} - {if data1.data2}{data1.data2}{end}', log), '{level} - bar')
+  })
+
+  t.test('parseMessageFormat translates if / else statement to found property value and leave unmatched property key untouched', async t => {
+    const log = fastCopy(logData)
+    t.equal(internals.parseMessageFormat('{level} - {if data1.data2}{data1.data2} ({msg}){end}', log), '{level} - bar ({msg})')
+  })
+
+  t.test('parseMessageFormat removes unescaped if statements', async t => {
```
jsumners, line 245: I believe you mean non-terminated instead of unescaped?

test/lib/utils.internals.test.js
```diff
@@ -222,3 +222,65 @@ tap.test('#getPropertyValue', t => {
...
+    const log = fastCopy(logData)
+    t.equal(internals.parseMessageFormat('{level} - {if data1.data2}{data1.data2}', log), '{level} - {data1.data2}')
+  })
+
+  t.test('parseMessageFormat removes unescaped end statements', async t => {
```
jsumners, line 250: I think "floating" is a better term than "unescaped" here.

jsumners, line 250: What happens if the format string is?:

`{level} - {if foo}something{end}{end}`

(i.e. someone tried to add an unsupported nested condition but failed to add the leading `if`)

lib/utils.js
```diff
@@ -766,3 +772,32 @@ function handleCustomlevelNamesOpts (cLevels) {
...
+ * {if key}{key}{end} statements and returns everything between if and else blocks if the key provided
+ * was found in log.
+ *
+ * @param {string} messageFormat A format string or function that defines how the
+ *  logged message should be conditionally formatted, e.g. `'{if level}{level}{end} - {if req.id}{req.id}{end}'`.
+ * @param {object} log The log object to be modified.
+ *
+ * @returns {string} The parsed messageFormat.
+ */
+function parseMessageFormat (messageFormat, log) {
```
jsumners, line 787: I think this would be a better name:

```suggestion
function interpretConditionals (messageFormat, log) {
```

jsumners (changes requested):

timlohse1104:
Hey @jsumners thanks for the code review. I will look into it after work. Please excuse my poor english var and function wording. Will edit.

test/lib/utils.internals.test.js
timlohse1104: Added further test scenarios.

test/lib/utils.internals.test.js
timlohse1104: Changed wording.

test/lib/utils.internals.test.js
timlohse1104: Changed wording.

timlohse1104: The additional `{end}` within the if / end statement will be removed. Added test scenario for this.

lib/utils.js
timlohse1104: Renamed accordingly.

lib/utils.js
```diff
@@ -766,3 +772,34 @@ function handleCustomlevelNamesOpts (cLevels) {
...
+ *
+ * @returns {string} The parsed messageFormat.
+ */
+function interpretConditionals (messageFormat, log) {
+  messageFormat = messageFormat.replace(/{if (.*?)}(.*?){end}/g, replacer)
+
+  // Remove unescaped if blocks
+  messageFormat = messageFormat.replace(/{if (.*?)}/g, '')
+  // Remove unescaped end blocks
+  messageFormat = messageFormat.replace(/{end}/g, '')
```
jsumners, line 793:
```suggestion
  // Remove non-terminated if blocks
  messageFormat = messageFormat.replace(/{if (.*?)}/g, '')
  // Remove floating end blocks
  messageFormat = messageFormat.replace(/{end}/g, '')
```

timlohse1104: Changed accordingly

mcollina (approved):
lgtm

jsumners (approved):
Looks good to me.

PR #443, opened 2023-07-24
Remove coveralls

jsumners:
We have 100% coverage and tap will error out if we drop below it. We don't need Coveralls's comment spam to tell us coverage dropped.


mcollina (approved):
lgtm

Issue #444, opened 2023-07-25
New option to hide level and timestamp at the beginning of a line


timlohse1104:
In my testing with `--messageFormat` option i came across the question if it would be possible to mirror standard NestJS logs like this:

`[Nest] 45816  - 23.07.2023, 22:09:14     LOG [NestFactory] Starting Nest application... +59ms`

I tried to rebuild it with '--messageFormat' string / function but could not remove the `timestamp`and `level`in the beginning of every log.

I probably overlooked something. But if thats not the case i would like to add another option 'hideMetadata' (name to be discussed). 

I would add logic into https://github.com/pinojs/pino-pretty/blob/master/index.js line 135 and following.

What do you think?

Would also contribute to this closed issue: https://github.com/pinojs/pino-pretty/issues/437

timlohse1104:
Please consider my pr https://github.com/pinojs/pino-pretty/pull/445 .

vladshcherbin:
try

```js
{
  include: ''
}
```

or

```js
{
  ignore: 'level,time'
}
```

timlohse1104:
Thanks @vladshcherbin that's avalid option. No need for a further parameter. This issue can be closed.

timlohse1104:
For anyone interested in a config to recreate NestJS logs see the below .pino-pretty.rc.

{
  "ignore": "pid,hostname,level,time",
  "translateTime": false,
  "singleLine": false,
  "colorize": true,
  "hideObject": true, 
  "messageFormat": "[Nest] {pid}  - {if req.id} ({req.id}) - {end}{time} {level} {if context}[{context}] {end} {msg}"
}

PR #445, opened 2023-07-25
Add support to hide metadata in log line

timlohse1104:
See https://github.com/pinojs/pino-pretty/issues/444
---

With option `--hideMetadata` it is possible to transform a basic log line like this:

`[25.7.2023, 15:57:32] INFO: [Nest] 25.7.2023, 15:57:32 [RoutesResolver] | OcrController {/ocr}: {"context":"RoutesResolver"}`

into a simulated NestJS log line like this (provided you use the `messageFormat` option):

`[Nest] 25.7.2023, 15:57:32 [RoutesResolver] | OcrController {/ocr}: {"context":"RoutesResolver"}`

`--hideMetadata` option effectively hides `epoch`, `level`and other metadata like `reqId` from beeing printed in front of the actual log message.

### DIFF
--- a/index.js
+++ b/index.js
@@ -49,7 +49,8 @@ const defaultOptions = {
   hideObject: false,
   ignore: 'hostname',
   include: undefined,
-  singleLine: false
+  singleLine: false,
+  hideMetadata: false
 }
 
 function prettyFactory (options) {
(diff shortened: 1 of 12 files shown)


mcollina (approved):
lgtm

index.js
```diff
@@ -127,39 +128,42 @@ function prettyFactory (options) {
...
-    if (opts.levelFirst && prettifiedLevel) {
-      line = `${prettifiedLevel}`
-    }
 
-    if (prettifiedTime && line === '') {
-      line = `${prettifiedTime}`
-    } else if (prettifiedTime) {
-      line = `${line} ${prettifiedTime}`
-    }
+    if (!opts.hideMetadata) {
```
jsumners, line 133: Can this be done in some other way than continually nesting the core code of this function further and further to the right?

jsumners (changes requested):

index.js
timlohse1104: Do you have something in mind. I could see some options. One with higher refactoring effort like using strategy pattern (https://refactoring.guru/design-patterns/strategy) and some with lower effort like using switch statements/  lookup tables.

jsumners: I think a simple approach would be to move the now optional code path into a function.

timlohse1104: Added an approach

index.js
```diff
@@ -206,6 +178,46 @@ function prettyFactory (options) {
     }
 
     return line
+
+    function createLineMetadata () {
```
jsumners, line 182: Does this function need to be defined as an inner function of `prettyFactory`? Or can we put it in `lib/utils`?

timlohse1104: We could put it in `lib/utils` but would have to pass many parameters like `log`, `colorizer`, `levelKey` etc. Think i would prefer to leave it in index.js. What do you think about the naming?

jsumners: My concern is legibility and maintainability. The inlined function is easier to read than the nested conditional version, but it is still a lot of code deeply nested within another function.

I think the function name is fine.

jsumners: @timlohse1104 I think you'll find it much easier to accommodate the ask now that #453 has been merged. Unfortunately, it'll be easier to start with a fresh PR. Rebasing this one may be too difficult.

If you have any questions, don't hesitate to ask.

PR #449, opened 2023-08-28
Re-org code

jsumners:
The impetus for this PR was me starting to work on a suggestion PR for #445. When I got into the code, I decided that we have long ago outgrown the structure of the codebase as I created it when splitting this module out of the core Pino module. Basically, it has grown in a difficult to navigate mess with everything crammed into a few large files.

I have more that I intend to do, but I think this PR has become large enough as it is. So I'm halting my work until we make a decision on this (🤞 that means it's merged).

-----

This PR is not yet complete as of commit https://github.com/pinojs/pino-pretty/commit/065fc998391666bb2671034b855493418f30a34d. I just want to put it up to show the following:

As of that commit, the "utils" file has been split out into a directory, `lib/utils/`. This directory has one file per utility function and a corresponding test file for each utility function file. It also updates the `tap` configuration to include a coverage map. The result is that we are not actually covering at 100% as we assumed:

<details>
<summary>Coverage Report</summary>

```sh
❯ tap
Suites:   ​27 passed​, ​27 of 27 completed​
Asserts:  ​​​453 passed​, ​of 453​
ERROR: Coverage for lines (95.48%) does not meet global threshold (100%)
ERROR: Coverage for functions (93.75%) does not meet global threshold (100%)
ERROR: Coverage for branches (85.77%) does not meet global threshold (100%)
ERROR: Coverage for statements (93.29%) does not meet global threshold (100%)
-------------------------------------|---------|----------|---------|---------|-------------------
File                                 | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s 
-------------------------------------|---------|----------|---------|---------|-------------------
All files                            |   93.29 |    85.77 |   93.75 |   95.48 |                   
 lib                                 |     100 |    90.62 |     100 |     100 |                   
  colors.js                          |     100 |    90.62 |     100 |     100 | 47-48,59          
 lib/utils                           |   92.19 |    84.97 |   91.66 |   94.71 |                   
  build-safe-sonic-boom.js           |   93.75 |    66.66 |   66.66 |   93.75 | 21                
  create-date.js                     |     100 |      100 |     100 |     100 |                   
  delete-log-property.js             |     100 |       80 |     100 |     100 | 24                
  filter-log.js                      |     100 |      100 |     100 |     100 |                   
  format-time.js                     |   95.23 |    93.33 |     100 |   95.23 | 45                
  get-property-value.js              |     100 |      100 |     100 |     100 |                   
[...]

PR #452, opened 2023-08-30
Refactor functions to use context object

jsumners:
This PR refactors all of the prettify functions to use a context object. This removes the need to keep track of all of the parsed options and other items such functions need in order to perform their work. I believe this will make the code base easier to maintain going forward, and will make PRs like #445 easier to implement.

----

It's not clear to me why the following block of code in https://github.com/pinojs/pino-pretty/commit/770f56a44e0d4a30e2809c379ded5647aeba43df stops working when refactoring it to use the context object:

https://github.com/pinojs/pino-pretty/blob/fa386a964cce7d7c5cbeb4ce5bb8d28785001acd/index.js#L117-L122

The short of it is that the following test starts failing:

https://github.com/pinojs/pino-pretty/blob/fa386a964cce7d7c5cbeb4ce5bb8d28785001acd/test/basic.test.js#L642-L664

It starts failing because it ultimately invokes `Number("info")`, resulting in `NaN`, on line 119 of the current `master` code because `this.customLevels` defaults to an empty object. This is mysterious to me because the same default is present on the master branch.

Regardless, unrolling the conditionals is far easier to read and reason through. But I'm still likely to revisit this block before considering this PR ready.

Update: I thought it might be the same thing as this:
https://github.com/pinojs/pino-pretty/blob/a74b7b1d37ff227f04e5aea8d11a1f33bf2d106b/lib/pretty.js#L88-L91

But it does not seem to be. I'm going to leave the unrolled block in place. I really don't like the conditionals nested in ternaries to begin with.

### DIFF
--- a/index.js
+++ b/index.js
@@ -65,8 +65,9 @@ const pretty = require('./lib/pretty')
  * lines.
  * @property {string} [timestampKey='time'] Defines the key in incoming logs
  * that contains the timestamp of the log, if present.
- * @property {boolean} [translateTime=true] When true, will translate a
- * JavaScript date integer into a human-readable string.
+ * @property {boolean|string} [translateTime=true] When true, will translate a
+ * JavaScript date integer into a human-readable string. If set to a string,
+ * it must be a format string.
  * @property {boolean} [useOnlyCustomProps=true] When true, only custom levels
  * and colors will be used if they have been provided.
  */
--- a/index.js
+++ b/index.js
@@ -4,36 +4,17 @@ const { isColorSupported } = require('colorette')
 const pump = require('pump')
 const { Transform } = require('readable-stream')
 const abstractTransport = require('pino-abstract-transport')
-const sjs = require('secure-json-parse')
 const colors = require('./lib/colors')
 const {
   ERROR_LIKE_KEYS,
   MESSAGE_KEY,
-  TIMESTAMP_KEY,
-  LEVELS,
-  LEVEL_KEY,
-  LEVEL_NAMES
+  TIMESTAMP_KEY
 } = require('./lib/constants')
 const {
-  isObject,
-  prettifyErrorLog,
-  prettifyLevel,
-  prettifyMessage,
-  prettifyMetadata,
-  prettifyObject,
-  prettifyTime,
   buildSafeSonicBoom,
-  filterLog,
   parseFactoryOptions
 } = require('./lib/utils')
-
-const jsonParser = input => {
-  try {
-    return { value: sjs.parse(input, { protoAction: 'remove' }) }
-  } catch (err) {
-    return { err }
-  }
-}
+const pretty = require('./lib/pretty')
 
 /**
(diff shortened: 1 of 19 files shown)


jsumners:
@mcollina in order to avoid the situation in #449 that led to having to create #453, once this and #453 are approved I will be merging in the following order:

1. #452 
2. #450 
3. #452

It will negate the approvals on those issues as I move through them, but nothing will be changing other than everything coming together. I broke this up into multiple PRs in an effort to make it a least somewhat easier to review.

index.js
```diff
@@ -37,42 +26,50 @@ const jsonParser = input => {
  * @property {boolean} [colorizeObjects=true] Apply coloring to rendered objects
  * when coloring is enabled.
  * @property {boolean} [crlf=false] End lines with `\r\n` instead of `\n`.
+ * @property {string|null} [customColors=null] A comma separated list of colors
```
jsumners, line 29: This doc block got reorganized to order the properties in alphabetical order. No real change here.

index.js
```diff
@@ -84,22 +81,25 @@ const defaultOptions = {
   colorize: isColorSupported,
   colorizeObjects: true,
   crlf: false,
+  customColors: null,
```
jsumners, line 84: Same as with the doc block. Properties were reorganized to be in alphabetical order.

index.js
```diff
@@ -110,168 +110,8 @@ const defaultOptions = {
...
-      if (singleLine && !/^\s$/.test(prettifiedObject)) {
-        line += ' '
-      }
-      line += prettifiedObject
-    }
-
-    return line
-  }
+  const context = parseFactoryOptions(Object.assign({}, defaultOptions, options))
+  return pretty.bind({ ...context, context })
```
jsumners, line 114: All of the implementation code was moved to `lib/pretty.js`.

lib/pretty.js
jsumners, line 1: New file that contains the original prettification code from the root `index.js`.

lib/pretty.js
```diff
@@ -0,0 +1,169 @@
...
+    }
+    if (!minimum) {
+      minimum = typeof this.minimumLevel === 'string'
+        ? LEVEL_NAMES[this.minimumLevel]
+        : LEVEL_NAMES[LEVELS[this.minimumLevel].toLowerCase()]
+    }
+
+    const level = log[this.levelKey === undefined ? LEVEL_KEY : this.levelKey]
+    if (level < minimum) return
+  }
```
jsumners, line 76: This is a change from the original code. See the PR description for details on this block.

lib/utils/parse-factory-options.js
jsumners, line 1: This is a new file that contains the original code from `prettyFactory` that parsed the passed in options. This one parses the options and returns a standardized context object that we can pass around to the various prettifier functions.

jsumners:
@mcollina when you get a chance, please review this one.

mcollina (approved):
great work much needed!

PR #467, opened 2023-10-04
Manually require and export all util scripts

### LINKED ISSUE #466: Test files included in release tarball and `module.exports` based on content of directory are messing with `esbuild` bundler
Release version 10.2.2 contains the following code which dynamically processes all JavaScript files in the same directory, supposedly excluding test files:

https://github.com/pinojs/pino-pretty/blob/9fb73303d48162e3332f0305143e159f98566d10/lib/utils/index.js#L6-L17

However, not all bundlers can handle this code. ESBuild for example describes their capability of bundling glob-style imports [here](https://esbuild.github.io/api/#glob). Since 

> Each non-string expression in the string concatenation chain becomes a wildcard in a glob pattern

it is unable to detect that `*.test.js` files are not to be bundled.

This leads to errors such as:

```
✘ [ERROR] Could not resolve "tap"

    node_modules/pino-pretty/lib/utils/join-lines-with-indentation.test.js:3:20:
      3 │ const tap = require('tap')
        ╵                     ~~~~~
```

This is easily reproducable using the following minimal setup:

<details>

`package.json`:

```json
{
  "main": "index.js",
  "scripts": {
    "esbuild": "esbuild index.js --bundle --platform=node --outfile=bundle.js"
  },
  "dependencies": {
    "esbuild": "0.19.4",
    "pino-pretty": "10.2.2"
  }
}
```

`index.js`:

```js
const pinoPretty = require('pino-pretty');
```

Running `find node_modules/pino-pretty -name "*.test.js" -delete` solves the issue.

</details>

Would it be possible to not ship `.test.js` files with pino-pretty, e.g. by removing them as part your build and release pipeline?

marvinruder:
* Fixes #466

### DIFF
--- a/lib/utils/index.js
+++ b/lib/utils/index.js
@@ -1,19 +1,28 @@
 'use strict'
 
-const { readdirSync } = require('fs')
-const { basename } = require('path')
-
-const files = readdirSync(__dirname).filter(f => f.endsWith('.test.js') === false)
-for (const file of files) {
-  if (file === 'index.js') continue
-
-  const kebabName = basename(file, '.js')
-  const snakeName = kebabName.split('-').map((part, idx) => {
-    if (idx === 0) return part
-    return part[0].toUpperCase() + part.slice(1)
-  }).join('')
-
-  module.exports[snakeName] = require(`./${kebabName}.js`)
+module.exports = {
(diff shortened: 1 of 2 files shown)


jsumners (approved):
Looks good to me.

PR #471, opened 2023-10-11
fix: exclude files from npm package

viceice:
Those files shouldn't be inside the npm package. Otherwise they can cause side effects.

https://github.com/containerbase/base/actions/runs/6480142703/job/17595082781?pr=1562

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,10 @@
+/.github/
+/docs/
+/help/
+/test/
+.taprc.yaml
+/benchmark.js
+/coverage-map.js
+/tsconfig.json
+
+*.test.js


jsumners:
Thank you for your interest in this topic. Please see the discussion in https://github.com/fastify/skeleton/issues/42#issuecomment-1208090783 for my reasoning against such a change.

viceice:
but the test files are a blocker for upgrade since you moved them to the lib folder

jsumners:
I don't know how that is the case.

viceice:
OK, the problem was fixed by
- #467

Issue #429, opened 2023-06-09
RC config file support for programmatic usage


adamhamlin:
Best I can tell, the rc files are only loaded for CLI usage (i.e., going through `bin.js`). Any plans to extend that to programmatic usage?

https://github.com/pinojs/pino-pretty/blob/849f0fc907adc3df30e7290b7fc091a055a30e51/bin.js#L23-L32

jsumners:
There are not any plans to do so. In my opinion, I think the readme is rather clear that this module is not meant for production use cases and adding such support would make it easier to use in unsupported contexts. The rc file support that is currently included is provided for CLI usage where providing options on each invocation is quite cumbersome otherwise. This is a common feature in CLI applications (pino-pretty's primary purpose).

I would say that if loading rc files is desired in programmatic usage, then it is the responsibility of the app.

adamhamlin:
Better late than never. Opened https://github.com/pinojs/pino-pretty/pull/475

PR #475, opened 2023-10-21
chore: Remove `config` from README programmatic usage options

adamhamlin:
Removes the `config` option from the README's programmatic options. The `config` option is only relevant to CLI usage.

A possible related enhancement could be to error for any "unknown" options (for example, passing `config` for programmatic usage is silently ignored at the moment).

Resolves https://github.com/pinojs/pino-pretty/issues/429


mcollina (approved):
lgtm

PR #484, opened 2023-12-27
Add build function export in type define file.

jaredhan418:


jaredhan418:
截屏2023-12-27 16 40 17

Then, user can use build function with TS type defined

mcollina (approved):
lgtm

PR #487, opened 2024-01-18
Support objects in options.customColors

kindagonzo:
Documentation notes that options.customColors can be either a string or object but objects are not currently supported in the code.
From index.d.ts:
  /**
   * Change the level colors to an user custom preset.
   *
   * Can be a CSV string in 'level_name:color_value' format or an object.
   * Also supports 'default' as level_name for fallback color.
   *
   * @example ( CSV ) customColors: 'info:white,some_level:red'
   * @example ( Object ) customColors: { info: 'white', some_level: 'red' }
   */
  customColors?: string|object;

This change permits either strings or objects to be used.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -150,3 +150,6 @@ profile-*
 
 # Run Configuration
 test/.tmp*
+
+# Local History
+.history
--- a/lib/utils/parse-factory-options.js
+++ b/lib/utils/parse-factory-options.js
@@ -87,23 +87,39 @@ function parseFactoryOptions (options) {
 
   let customColors
   if (options.customColors) {
-    customColors = options.customColors.split(',').reduce((agg, value) => {
-      const [level, color] = value.split(':')
-
-      const condition = useOnlyCustomProps
-        ? options.customLevels
-        : customLevelNames[level] !== undefined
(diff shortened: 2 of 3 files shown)


mcollina (changes requested):
Thanks for opening a PR! Can you please add a unit test?

kindagonzo:
> Thanks for opening a PR! Can you please add a unit test?

Sure. I checked around in the test folder but I can't seem to find any tests referencing that function. Where would you like me to add a test?

mcollina:
You can test it via `prettyFactory` (https://github.com/pinojs/pino-pretty/blob/9d89a596f492eb2efee1632f54be9bc9b93356f4/test/basic.test.js#L526) or add a new file altogether.

kindagonzo:
Can you help me out? I added the test and can run the suite fine with npm run test or npm run ci:
PinoPretty Success

However, the pre-commit hook fails:
PinoPretty Fail

Any idea why?

kindagonzo:
It seems to be a transient issue. I ran it a few times and was finally able to commit.

mcollina (approved):
lgtm

Issue #456, opened 2023-09-08
Regression with levelKey handling in version 10


paulish:
When pino-pretty is used together with elastic formatting the logging level has `log.level` key in the log object. I found no way to make pino-pretty work correctly with this level key.

In version 9 pino-pretty allowed dot in key names without escaping while in version 10 we need to escape it with slashes.
This would be ok but this escaping breaks excluding this key from the message string.

Example (uncomment levelKey with/without escaping to see different havaviours):

```javascript
const pino = require('pino');
const pretty = require('pino-pretty');
const ecsFormat = require('@elastic/ecs-pino-format');

const options = ecsFormat({});
options.level = 'debug';
const logger = pino(options, pretty({
    colorize: true,
    singleLine: true,
    levelKey: 'log\\.level', // if levelKey dot is escaped then log.level appears at the console output
    // levelKey: 'log.level', // if levelKey dot is not escaped then level is not showing between message
    timestampKey: '@timestamp',
    messageKey: 'message',
    ignore: 'service,process,host,ecs'
}));

logger.info('Starting');
logger.debug({
    extraParam: '123',
    message: 'hello hello'
});
```

The example output is 
![image](https://github.com/pinojs/pino-pretty/assets/1762032/9baede30-d9d8-4fc9-b8e7-48b51fba4c6c)

As you can see the output contains unwanted `"log.level":"info"` and `"log.level":"debug"`.

If I remove dot escaping then the output changes to
![image](https://github.com/pinojs/pino-pretty/assets/1762032/7adc133f-250e-4eef-a3fd-2b6c8961347c)

Ths time it has no colored `INFO` and `DEBUG` levels.

mcollina:
Thanks for reporting! Would you like to send a Pull Request to address this issue? Remember to add unit tests.

eliw00d:
@mcollina I have created a PR with a fix for this.

PR #490, opened 2024-01-24
fix: regression with showing standard keys in output object if they contain escaped dot

paulish:
fixes issue https://github.com/pinojs/pino-pretty/issues/456


jsumners (changes requested):
1. Missing unit tests.
2. Current tests are failing.

ademarsj:
It seems this has already been fixed by another Pull Request(#538).
Maybe discard this PR?

Issue #489, opened 2024-01-21
LevelName whitespace alignment issue


JohnPolansky:
I'm using the following packages:
```
    "pino": "^8.17.2",
    "pino-pretty": "^10.3.1",
```

When I print out any messages using the `colorize()` option it appears that the message alignment is off
![image](https://github.com/pinojs/pino-pretty/assets/8783582/fde9d9b5-7b1e-4c9d-9565-bd28e8636222)

As you can see the INFO/WARN are throwing off messages which makes it a bit harder to read log messages. After searching various older posts I found a reference to this same issue back in 2021 that indicated this is an issue and it was fixed. Has something changed or am I missing a feature? https://github.com/pinojs/pino-pretty/pull/141

I also tried creating my own transport to `padEnd()` the log level, but this doesn't appear to work as below 
```
import pino from 'pino-pretty'
import { colorizerFactory } from 'pino-pretty'
const levelColorize = colorizerFactory(true)
const levelPrettifier = logLevel => `${levelColorize(logLevel).padEnd(5)}`

export default (opts) => {
    return pino({
        ...opts,
        messageFormat: (log, messageKey) => `hello ${log[messageKey]}`
        customPrettifiers: {
            level: levelPrettifier
        }
    })
}
```

Here is my full logger.js for review:
```
import pino from 'pino';

//Create a logging instance
export const logger = pino({
  enabled: process.env.LOG_ENABLED === undefined ? true : process.env.LOG_ENABLED === 'true',
  transport: {
    // ...(process.env.LOG_PRETTY === true) && {'target': 'pino-pretty'},
    target: process.env.LOG_PRETTY === 'true' ? 'pino-pretty' : 'pino/file',
    // target: './pino-pretty-transport.js',
    options: {
      colorize: true,
      colorizeObjects: true,
      singleLine: true,
      ignore: 'pid,hostname',
    }
  },
  base: {},
  formatters: {
    level: (label) => {
      return { level: label };
    },
  },
  level: process.env.LOG_LEVEL || 'info',
  name: process.env.LOGGER_NAME,
  redact: {
    paths: ['token', 'password', 'uri'],
  },
  timestamp: pino.stdTimeFunctions.isoTime,
});
```
Thanks in advance for the help and the great tool.

mcollina:
Thanks for reporting. I don't exactly remember what happened.

@jsumners did you recall what happened on this one?

jsumners:
[...]

PR #493, opened 2024-02-22
feat: Add context to level customPrettifier

FoxxMD:
Currently the customPrettifier for `level` only returns the numeric level value (if `levelKey` is not used) which means that anyone who wants to modify value returned has to re-implement the levels pino-pretty already has access to as well as calculate the label to use, which pino-pretty also already does. 

This PR include both the "final" log label as well as the colorized output, if applicable, to `level` prettifier function to give users more control over level output.

It has the additional benefit of enabling users to handle their own level alignment (#489 #140) since they have access to the final string values for level. As example implementation for aligning with spaces:

```js
{
  customPrettifiers: {
        level: (out, label, colorized) => {
            // pad to fix alignment
            // assuming longest level is VERBOSE
            // and may be colorized
            const paddedLabel = label.padEnd(8)
            if(colorized !== label) {
                const padDiff = paddedLabel.length - label.length;
                return colorized.padEnd(colorized.length + padDiff);
            }
            return paddedLabel;
        }
    },
}

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -311,6 +311,9 @@ Additionally, `customPrettifiers` can be used to format the `time`, `hostname`,
     // on if the levelKey option is used or not.
     // By default this will be the same numerics as the Pino default:
     level: logLevel => `LEVEL: ${logLevel}`,
+    // Additionally, optional second and third arguments arguments can be provided
+    // to get the level label as a string and the colorized version (if `colorize: true`), respectively:
+    level: (logLevel, levelLabel, coloredLevelLabel) => `LEVEL: ${logLevel} LABEL: ${levelLabel} COLORIZED LABEL: ${coloredLevelLabel}`,
 
     // other prettifiers can be used for the other keys if needed, for example
     hostname: hostname => colorGreen(hostname),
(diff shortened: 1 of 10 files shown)


FoxxMD:
The only thing missing is that the Typescript typings for all customPrettifiers use `Record<string, (inputData: string | object) => string>` so TS reports errors when the end user uses `level: (out, label, colorized) => { ... ` and im not sure how to address this.

FoxxMD:
Modified typings...it should work now

FoxxMD:
Related to [#494](https://github.com/pinojs/pino-pretty/pull/494#issuecomment-1961758216) -- providing colors to all customPrettifier funcs would add _another_ argument to this function signature which I'm not super jazzed.

If I can get feedback on #494 and this cohesively -- or if ya'll would prefer I close both of these and open a new PR with everything combined -- let me know. But I'm thinking:

Change signature for customPrettifier to be `(value: string | object, extras?: object)` so that all these additional variables can be destructured from the second arg to avoid requiring the user to know argument order for multiple signatures.

mcollina (approved):
lgtm

mcollina:
@jsumners could you take a look?

FoxxMD:
Which part @mcollina ? This PR or the changes I mentioned above? https://github.com/pinojs/pino-pretty/pull/493#issuecomment-1964469268

mcollina:
@FoxxMD sorry I don't understand the question.

FoxxMD:
Please...read my comment above and respond to it. https://github.com/pinojs/pino-pretty/pull/493#issuecomment-1964469268

For these changes and #494, after I opened these PRs, I found the signatures to be unwiedly

> providing colors to all customPrettifier funcs would add another argument

I built out another branch, [`additionalFunctionality`](https://github.com/FoxxMD/pino-pretty/tree/additionalFunctionality) **that includes functionally the same changes** as this and #494 but with a more normalized signature for both to address this.

Take a look at the [`customPrettier` section on the readme in the `additionalFunctionality` branch](https://github.com/FoxxMD/pino-pretty/tree/additionalFunctionality?tab=readme-ov-file&rgh-link-date=2024-02-26T15%3A49%3A46Z#options) or [the diff](https://github.com/FoxxMD/pino-pretty/commit/2676ffd24a308a2638158f38cf147163162e8022) to see what changes I made.

Per my comment above:

> **Can I get feedback on #494 and this cohesively -- or would ya'll prefer I close both of these and open a new PR with everything combined**

mcollina:
I missed that. Please send one PR thanks.

PR #494, opened 2024-02-23
feat: Provide colorette object to messageFormat

FoxxMD:
Enables users to use available colors based on `colorize` context of the pino-pretty instance.

Currently if a user wants to color their messages they need to create their own instances of Colorette and implement logic, outside of pino-pretty, to make sure the `useColors` option passed matches what was given to pino-pretty as `colorize`. However pino-pretty already has this object available...this PR extends `messageFormat` to add the existing Colorette instance as a fourth parameter so that the user doesn't need to do that extra work.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -356,9 +356,11 @@ This option can also be defined as a `function` with this prototype:
 
 ```js
 {
-  messageFormat: (log, messageKey, levelLabel) => {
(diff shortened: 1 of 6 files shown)


FoxxMD:
Question though...would it be acceptable to extend this functionality to `customPrettifiers` so that colors can be accessed from essentially all functions for customizing output?

mcollina (approved):
lgtm

PR #495, opened 2024-03-01
Providing additional functionality to customPrettifiers and messageFormat

FoxxMD:
This PR is combined changes from #493 and #494 (see these comments [1](https://github.com/pinojs/pino-pretty/pull/493#issuecomment-1964469268) [2](https://github.com/pinojs/pino-pretty/pull/493#issuecomment-1971957478)). Summary of changes:

## Extend/Normalize all "format" functions with `extras` object

All "format" functions -- `customPrettifiers`, `level`, and `messageFormat` -- now have an additional parameter, `extras` that is an object of `Record<string, any>`. This allows for additional data and future (optional) functionality to be provided to these functions without having to change function signature again.

This change is backward compatible but the examples in the documentation for these functions has been updated to include the existing, full, list of parameters that were available instead of just the first parameter. EX

Previous:

```js
{
  customPrettifiers: {
    caller: caller => `My Caller: ${caller}`
  }
}
```
```js
{
  messageFormat: (log, messageKey, levelLabel) => {
    return customized_message;
  }
}
```

New:
```js
{
  customPrettifiers: {
    caller: (caller, key, log, extras) => `My Caller: ${caller}`,
  }
}
```
```js
{
  messageFormat: (log, messageKey, levelLabel, extras) => {
    return `This is a custom message: ${log[messageKey]}`;
  }
}
```

## Provide Colorette `colors` to all "format" functions #494

Enables users to use available colors based on `colorize` context of the pino-pretty instance.

Currently if a user wants to color their messages they need to create their own instances of Colorette and implement logic, outside of pino-pretty, to make sure the `useColors` option passed matches what was given to pino-pretty as `colorize`. However pino-pretty already has this object available...this PR provides `colors` to the `extras` object for all format functions so that the user doesn't need to do that extra work.

```js
{
  customPrettifiers: {
    caller: (caller, key, log, { colors } ) => `${colors.greenBright(caller)}`,
  },
  messageFormat: (log, messageKey, levelLabel, { colors }) => {
    // do some log message customization
    //
    // `colors` is a Colorette object with colors enabled based on `colorize` option
    return `This is a ${color.red('colorized')}, custom message: ${log[messageKey]}`;
  }
}
```

## Add context to `level` customPrettifier #493

Currently the customPrettifier for `level` only returns the numeric level value (if `levelKey` is not used) which means that anyone who wants to modify value returned has to re-implement the levels pino-pretty already has access to as well as calculate the label to use, which pino-pretty also already does. 

This PR include both the "final" log label as well as the colorized output, if applicable, to `level` prettifier function to give users more control over level output. These values are included in the `extras` object:

```js
{
  customPrettifiers: {
  // level provides additional data in `extras`:
  // * label => derived level label string
  // * labelColorized => derived level label string with colorette colors applied based on customColors and whether colors are supported
  level: (logLevel, key, log, { label, labelColorized, colors }) => `LEVEL: ${logLevel} LABEL: ${levelLabel} COLORIZED LABEL: ${labelColorized}`,
  }
}
```

It has the additional benefit of enabling users to handle their own level alignment (#489 #140) since they have access to the final string values for level. As example implementation for aligning with spaces:

```js
{
  customPrettifiers: {
        level: (logLevel, key, log, { label, labelColorized }) => {
            // pad to fix alignment
            // assuming longest level is VERBOSE
            // and may be colorized
            const paddedLabel = label.padEnd(8)
            if(labelColorized !== label) {
                const padDiff = paddedLabel.length - label.length;
                return labelColorized.padEnd(labelColorized.length + padDiff);
            }
            return paddedLabel;
        }
    },
}
```

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -297,8 +297,23 @@ const prettifyQuery = value => {
 }
 ```
 
+All prettifiers use this function signature:
+
+```js
+['logObjKey']: (output, keyName, logObj, extras) => string
+```
+
+* `logObjKey` - name of the key of the property in the log object that should have this function applied to it
+* `output` - the value of the property in the log object
+* `keyName` - the name of the property (useful for `level` and `message` when `levelKey` or `messageKey` is used)
+* `logObj` - the full log object, for context
+* `extras` - an object containing **additional** data/functions created in the context of this pino-pretty logger or specific to the key (see `level` prettifying below)
+  * All `extras` objects contain `colors` which is a [Colorette](https://github.com/jorgebucaran/colorette?tab=readme-ov-file#supported-colors) object containing color functions. Colors are enabled based on `colorize` provided to pino-pretty or `colorette.isColorSupported` if `colorize` was not provided.
+
 Additionally, `customPrettifiers` can be used to format the `time`, `hostname`,
-`pid`, `name`, `caller` and `level` outputs:
+`pid`, `name`, `caller` and `level` outputs AS WELL AS any arbitrary key-value that exists on a given log object.
+
+An example usage of `customPrettifiers` using all parameters from the function signature:
 
 ```js
 {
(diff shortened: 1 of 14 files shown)


FoxxMD:
Please hold off...there's a bug somewhere that is showing up in my app but not in tests.

FoxxMD:
Fixed :+1:

mcollina (approved):
lgtm

mcollina:
@jsumners I think it's safer to bump a major on this one.

FoxxMD:
@mcollina thank you for the effort on getting pino-roll updated! Any chance this can move forward too? :)

PR #498, opened 2024-03-13
Update help.txt: Fix program name

glensc:
> $ pino pretty
>`pino` cli has been removed. Use `pino-pretty` cli instead.
>
> See: https://github.com/pinojs/pino-pretty


mcollina (approved):
lgtm

PR #508, opened 2024-04-29
feat: automatically consume pino config when using pino@>=8.21.0

10xLaCroixDrinker:
This update consumes the new feature introduced in pinojs/pino#1930

I opted to listen directly for the `PINO_CONFIG` message here instead of using the `expectPinoConfig` option in pino-abstract-transport in order to support use outside `pino.transport()` and maintain maximum backward compatibility.


mcollina (approved):
lgtm

PR #514, opened 2024-05-26
Fix: add missing useOnlyCustomProps options to type definitions

pcdevil:
Hello,

I started using pino-pretty in a TypeScript project and noticed the `useOnlyCustomProps` option is not present in the type definitions module. The JSDoc lists it at [L71-72](https://github.com/pinojs/pino-pretty/blob/77e85fdcf3413b5064c5258ea3b858eac230340f/index.js#L71-L72), so I assume it was a copy-paste error at some time.

This pull request addresses this discrepancy by copying the [readme's description](https://github.com/pinojs/pino-pretty/blob/77e85fdcf3413b5064c5258ea3b858eac230340f/Readme.md?plain=1#L80)!

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -210,6 +210,12 @@ interface PrettyOptions_ {
    * @example ( Object ) customColors: { info: 'white', some_level: 'red' }
    */
   customColors?: string|object;
+  /**
+   * Only use custom levels and colors (if provided); else fallback to default levels and colors.
+   *
+   * @default true
+   */
+  useOnlyCustomProps?: boolean;
 }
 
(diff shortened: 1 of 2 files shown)


mcollina (commented):
Thanks for opening a PR! Can you please add a unit test? We use tsd for types.

pcdevil:
@mcollina absolutely! I'm not familiar with tsd, but based on the work on the test suite it seemed adding the property suffice for the test expectation. please let me know if it requires more work 🙇🏼 

_disclaimer: I opted to amend the existing commit to not clutter the git history - this way the change and the test are located in a single commit. I hope you don't mind!_

PR #522, opened 2024-07-22
Type fixes

### LINKED ISSUE #521: TypeScript: "no call signatures" when using ESM
When configuring a project for ESM (`"type": "module"` in package.json, and tsconfig.json using `"moduleResolution": "node16",   "module": "node16"`), if I try to import PinoPretty as `import PinoPretty from 'pino-pretty'` and call it, I get TypeScript error:

```
src/node.ts:23:48 - error TS2349: This expression is not callable.
  Type 'typeof import("node_modules/pino-pretty/index")' has no call signatures.

23       PinoPretty({ colorize: true })

Found 1 error in src/node.ts:23
```

If I instead try a named import (`import { PinoPretty } from 'pino-pretty'`), then TypeScript compiles, but Node.js throws an error at runtime:

```
import { PinoPretty } from 'pino-pretty';
         ^^^^^^^^^^
SyntaxError: Named export 'PinoPretty' not found. The requested module 'pino-pretty' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from 'pino-pretty';
const { PinoPretty } = pkg;
```

I noticed that [Are the Types Wrong](https://arethetypeswrong.github.io/?p=pino-pretty%4011.2.1) is reporting a ["Missing export =" error](https://github.com/arethetypeswrong/arethetypeswrong.github.io/blob/main/docs/problems/MissingExportEquals.md) for pino-pretty. I believe that this relates to the error I'm seeing.

joshkel:
This is a proposal for fixing #521. It works in the handful of local tests that I've done, but it probably needs more testing. I can test further if this general approach is acceptable.

See individual commits for specific changes:

* (unrelated) Fix a few typos
* Change to use a `declare namespace` throughout, following the "main export is a function" [example from Are the Types Wrong](https://github.com/arethetypeswrong/arethetypeswrong.github.io/blob/main/docs/problems/MissingExportEquals.md)
* Add Are the Types Wrong as a build check
* Add a named export of `PinoPretty` - This was declared by the pre-existing types but was missing from the JS, causing runtime errors. It's easy to update the JS to match the intention of the pre-existing types.

Fixes #521


mcollina (approved):
lgtm

PR #538, opened 2024-11-04
fix: unescape \\ for levelKey

### LINKED ISSUE #456: Regression with levelKey handling in version 10

eliw00d:
closes #456

### DIFF
--- a/lib/pretty.js
+++ b/lib/pretty.js
@@ -147,11 +147,13 @@ function pretty (inputData) {
       this.messageKey,
       this.levelKey,
       this.timestampKey
-    ].filter(key => {
-      return typeof log[key] === 'string' ||
-        typeof log[key] === 'number' ||
-        typeof log[key] === 'boolean'
-    })
+    ]
+      .map((key) => key.replaceAll(/\\/g, ''))
+      .filter(key => {
+        return typeof log[key] === 'string' ||
+          typeof log[key] === 'number' ||
+          typeof log[key] === 'boolean'
+      })
     const prettifiedObject = prettifyObject({
       log,
       skipKeys,
--- a/test/basic.test.js
+++ b/test/basic.test.js
@@ -218,6 +218,22 @@ test('basic prettifier tests', (t) => {
     log.info({ msg: 'foo', bar: 'warn' })
   })
 
+  t.test('can use nested level keys', (t) => {
+    t.plan(1)
+    const pretty = prettyFactory({ levelKey: 'log\\.level' })
+    const log = pino({}, new Writable({
+      write (chunk, enc, cb) {
+        const formatted = pretty(chunk.toString())
+        t.equal(
+          formatted,
+          `[${formattedEpoch}] WARN (${pid}): foo\n`
+        )
+        cb()
+      }
+    }))
+    log.info({ msg: 'foo', 'log.level': 'warn' })
+  })
+
   t.test('can use a customPrettifier on default level output', (t) => {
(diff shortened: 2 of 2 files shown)


lib/pretty.js
```diff
@@ -147,11 +147,13 @@ function pretty (inputData) {
       this.messageKey,
       this.levelKey,
       this.timestampKey
-    ].filter(key => {
-      return typeof log[key] === 'string' ||
-        typeof log[key] === 'number' ||
-        typeof log[key] === 'boolean'
-    })
+    ]
+      .map((key) => key.replaceAll(/\\/g, ''))
```
eliw00d, line 151: This strips off the `\\` in `log\\.level` so that you get the correct key to ignore.

test/basic.test.js
```diff
@@ -218,6 +218,22 @@ test('basic prettifier tests', (t) => {
...
 
+  t.test('can use nested level keys', (t) => {
+    t.plan(1)
+    const pretty = prettyFactory({ levelKey: 'log\\.level' })
+    const log = pino({}, new Writable({
+      write (chunk, enc, cb) {
+        const formatted = pretty(chunk.toString())
+        t.equal(
+          formatted,
+          `[${formattedEpoch}] WARN (${pid}): foo\n`
```
eliw00d, line 229: Without this fix, you would get `log.level`  underneath in the prettified object.

mcollina (approved):
lgtm

jsumners:
Yes, we should drop 14. It'll be a new major, though.

eliw00d:
> Yes, we should drop 14. It'll be a new major, though.

Should I make that change in this PR? For example:

```yml
      matrix:
        node-version: [18.17, 20, 22]
        os: [ubuntu-latest]
        pino-version: [^9.0.0]
```

mcollina:
@eliw00d send a fresh PR.

eliw00d:
> @eliw00d send a fresh PR.

[Done](https://github.com/pinojs/pino-pretty/pull/540).

eliw00d:
@jsumners @mcollina Should be good to go now! 🎉

mcollina (approved):
lgtm

jsumners (approved):

PR #540, opened 2024-11-06
Drop support for Node.js 14 & 16

eliw00d:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "pino-pretty",
-  "version": "11.3.0",
+  "version": "12.0.0",
```
jsumners, line 3: Please do not change this. The releaser will do it.

jsumners (changes requested):

package.json
eliw00d: Gotcha.

jsumners (approved):

Issue #541, opened 2024-11-07
Dropping `readable-stream` polyfill


benmccann:
It looks like the CI tests Node 18.17+, so I guess that might be the minimum version supported by this library. Once that's raised to Node 18.19+ it would be nice to drop the `readable-stream` dependency. I'm guessing that would be a change in supported Node versions and would be considered a breaking change, so might have to wait until the next major.

kibertoad:
Node 18 goes EOL on 2025-04-30, so personally I don't see a big problem with bumping minimum version to Node 20 in the nearby future (in a semver major, of course).
fastify has already done that.

as a bonus, node:test could be used instead of tap then

mcollina:
I'm +1 on removing, send a PR.

PR #542, opened 2024-11-08
chore: replace readable-stream with built-in stream

benmccann:
closes https://github.com/pinojs/pino-pretty/issues/541

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,7 +40,7 @@ jobs:
       contents: read
     strategy:
       matrix:
-        node-version: [18.19, 20, 22]
+        node-version: [20, 22]
         os: [ubuntu-latest]
         pino-version: [^9.0.0]
     steps:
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -40,7 +40,7 @@ jobs:
       contents: read
     strategy:
       matrix:
-        node-version: [18.17, 20, 22]
+        node-version: [18.19, 20, 22]
```
kibertoad, line 43: isn't that technically a semver major? I'd rather use the opportunity to drop node 18 from the list altogether

mcollina:
@kibertoad we just bumped the major, I don't think we need another bump.

mcollina:
cc @jsumners

jsumners:
We probably should have done it then, but I'm really not opposed to version churn on this module. I consider this module non-essential, and as such don't think it matters too much how often it is released.

PR #545, opened 2024-11-19
Readme.md: remove the command prompt symbol

yegorich:
Remove "$" because otherwise, it will also be copied to the clipboard and the pasted command will produce an error.


mcollina (approved):
lgtm

PR #564, opened 2025-05-09
test: enable borp to move to node test

Puppo:
This PR adds borp to run tests and  enables the project to move to node test


mcollina (approved):
Lgtm

jsumners (approved):
Awesome. We don't need a `next` branch here because we already only test on 20 and 22.

PR #569, opened 2025-05-11
test: move crlf.test.js to node test

Puppo:
This PR moves `crlf.test.js` to node test

### DIFF
--- a/test/crlf.test.js
+++ b/test/crlf.test.js
@@ -2,7 +2,7 @@
 
 process.env.TZ = 'UTC'
 
-const { test } = require('tap')
+const { test } = require('node:test')
 const _prettyFactory = require('../').prettyFactory
 
 function prettyFactory (opts) {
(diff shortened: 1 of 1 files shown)


test/crlf.test.js
```diff
@@ -16,20 +16,18 @@ function prettyFactory (opts) {
 
 const logLine = '{"level":30,"time":1522431328992,"msg":"hello world","pid":42,"hostname":"foo"}\n'
 
-test('crlf', (t) => {
-  t.test('uses LF by default', (t) => {
+test('crlf', async (t) => {
```
mcollina, line 19: Can you use describe instead?

mcollina (changes requested):

test/crlf.test.js
Puppo: @mcollina I can use whatever you prefer, but please align with @jsumners 🙂
In Pino, he asked me to avoid describe and use test.
So just let me know which way you prefer and I’ll use it.

mcollina: Sorry for the back-and-forth.

describe+test supports only, while the pattern used here does not.

cc @jsumners

jsumners: I think it's part of this? https://github.com/nodejs/node/pull/56664

Puppo: @mcollina Sorry if I ping you.
Is there any update on this one? So we can move forward with these PRs.

jsumners: I'm not going to belabor it. Use `describe`.

Puppo: Done

jsumners (approved):

PR #578, opened 2025-05-14
test: move filter-log.test.js to node test

Puppo:
This PR moves `filter-log.test.js` to node test


mcollina (approved):
LGTM! CI is passing and the test migration to node:test looks good.

PR #592, opened 2025-05-14
test: move prettify-message.test.js to node test

Puppo:
This PR moves `prettify-message.test.js` to node test


mcollina (approved):
LGTM! CI is passing and the test migration to node:test looks good.

PR #598, opened 2025-05-27
chore: upgrade dateformat

Uzlopak:
I saw, that dateformat is reported on npmgraph as being one major behind. This PR upgrades dateformat to the latest version


mcollina (approved):
lgtm

PR #614, opened 2025-08-07
fix: modify to run with @pinojs/dateformat

Uzlopak:


Uzlopak:
Basically SYS: is the same as GMT. I could just modify our dateformat to even avoid that `slice(4)` call.

mcollina (approved):
looks good!

Uzlopak:
No, there are no technical reasons. I made a mistake, by making a capital F instead of a lowercase f. I will fix it. I am just currently commuting... I provide you with a fix in fee hours

Uzlopak:
@jsumners
done

mcollina (approved):
lgtm once our dateformat method hits 1.0.0

lib/utils/format-time.js
```diff
@@ -46,21 +98,23 @@ function formatTime (epoch, translateTime = false) {
   }
 
   if (translateTime === true) {
-    return dateformat(instant, DATE_FORMAT_SIMPLE)
+    return simpleDateFormatter.format(instant)
```
jsumners, line 101: I'm confused as to why the `dateformat` function doesn't support the original format strings. We're having to do more work specific to date formatting in this module, which I don't think is a primary concern for it, with these changes.

Uzlopak:
Changing this PR to draft. Will get back to it some day.

lib/utils/format-time.js
jsumners: I certainly am not making that claim. I am trying to get the module incorporated. To do so, I want the incorporation to consist of replacing the module name in the `require` specifier only. I think this would be expected by anyone attempting to replace their usage of the original `dateformat` with the new one.

PR #635, opened 2025-11-22
fix: messageFormat print 0 value

gutenye:
If the value is 0, it won't print

Reproduce

```sh
echo {"a":0} | pino-pretty --messageFormat '{a}'
```

It outputs blank string

Expected outputs '0'


jsumners (changes requested):
Thank you. We need a test before this can be accepted.

mcollina (approved):
lgtm

PR #665, opened 2026-06-24
feat(cli): expose sonic-boom options

### LINKED ISSUE #664: Undocumented option: --sync
I wasted quite some time trying to figure out how to make a log pretty print with the correct order between non-pino lines and pino lines. The `--sync` option is not documented within `--help`. By accident, it works despite not being explicit for minimist…

Would be great to have it within `--help` and maybe explain it preventing out-of-order non-pino lines.

oleasteo:
Resolves #664

### DIFF
--- a/bin.js
+++ b/bin.js
@@ -42,8 +42,11 @@ if (cmd.h || cmd.help) {
 
   let opts = minimist(process.argv, {
     alias: {
+      // defaults to true (sonic-boom options default)
+      append: 'append',
       colorize: 'c',
       crlf: 'f',
+      destination: 'd',
       errorProps: 'e',
       levelFirst: 'l',
       minimumLevel: 'L',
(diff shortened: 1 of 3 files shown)


bin.js
```diff
@@ -42,8 +42,11 @@ if (cmd.h || cmd.help) {
 
   let opts = minimist(process.argv, {
     alias: {
+      // defaults to true (sonic-boom options default)
+      append: 'append',
```
mcollina, line 46: I don't think this is needed

bin.js
```diff
@@ -60,7 +64,8 @@ if (cmd.h || cmd.help) {
       ignore: 'i',
       include: 'I',
       hideObject: 'H',
-      singleLine: 'S'
+      singleLine: 'S',
+      sync: 'sync'
```
mcollina, line 68: I don't think this is needed

bin.js
oleasteo: True. They are not needed. However, I would prefer those for code quality reasons. I understand the call to minimist as a registration of possible program params. It's weird to me to have some params not being mentioned at all. The list would look *incomplete* to me here.
Your call, tho.

PR #676, opened 2026-08-02
fix: substitute conditional messageFormat values literally

Jaybhade:
### The problem

When a value is inserted into a `{if key}...{end}` block, `interpretConditionals` uses a string replacement:

```js
return value.replace(new RegExp('{' + key + '}', 'g'), propertyValue)
```

That line interpolates untrusted input into both halves of `String.prototype.replace`, which causes two distinct bugs.

**1. `$` sequences in the logged value are interpreted as replacement patterns.**

```js
const { prettyFactory } = require('pino-pretty')

const conditional = prettyFactory({ colorize: false, messageFormat: '{if msg}{msg}{end}' })
const plain = prettyFactory({ colorize: false, messageFormat: '{msg}' })

const log = (msg) => JSON.stringify({ level: 30, time: 1, msg })
```

| `msg` | `{if msg}{msg}{end}` | `{msg}` (expected) |
|---|---|---|
| `100$$` | `100$` | `100$$` |
| ``a $` b`` | `a b` | ``a $` b`` |
| `a $& b` | `a a $& b b` | `a $& b` |

`$$` collapses to a single `$`, ``$` `` and `$'` are replaced by the text around the match, and `$&` re-inserts the `{msg}` placeholder — which `prettifyMessage` then expands a second time, duplicating the value. Values like these turn up in practice in prices, shell snippets, and query strings.

**2. The key is compiled as a regular expression.**

The `.` that delimits a nested key matches any character, so an unrelated placeholder can be substituted with the wrong value:

```js
// log = { data1: { data2: 'bar' } }
interpretConditionals('{if data1.data2}{data1.data2} {data1Xdata2}{end}', log)
// actual:   'bar bar'
// expected: 'bar {data1Xdata2}'
```

And a key containing other special characters throws while formatting a log line:

```js
// log = { 'we(ird': 'value' }
interpretConditionals('{if we(ird}{we(ird}{end}', log)
// SyntaxError: Invalid regular expression: /{we(ird}/g: Unterminated group
```

### The fix

Match the placeholder and insert the value literally, via a string search and a function replacer. This is what the non-conditional path in `prettifyMessage` already does — it replaces through a function, which is why `{msg}` handles all of the cases above correctly today. The two paths now agree.

The `propertyValue && value.includes(key)` guard is unchanged, so which blocks are kept or dropped is not affected.

### Verification

Four tests added to `lib/utils/interpret-conditionals.test.js`, each failing before the change: `$` sequences preserved, every `$` replacement pattern round-tripped, a key with regex metacharacters no longer throwing, and the nested-key delimiter matched literally.

- `npm test` — 367 passed, 0 failed (363 before, plus the 4 new)
- `npm run ci` — `eslint` clean, `borp --check-coverage` passing, `tstyche` + `attw --pack` clean

### DIFF
--- a/lib/utils/interpret-conditionals.js
+++ b/lib/utils/interpret-conditionals.js
@@ -33,7 +33,11 @@ function interpretConditionals (messageFormat, log) {
   function replacer (_, key, value) {
     const propertyValue = getPropertyValue(log, key)
     if (propertyValue && value.includes(key)) {
-      return value.replace(new RegExp('{' + key + '}', 'g'), propertyValue)
+      /* The key is matched, and the value inserted, literally: a key may contain
+         characters that are special in a regular expression (e.g. the `.` used to
+         delimit nested keys), and a logged value may contain `$` sequences that
+         would otherwise be interpreted as replacement patterns. */
+      return value.replaceAll(`{${key}}`, () => propertyValue)
     } else {
       return ''
     }
--- a/lib/utils/interpret-conditionals.test.js
+++ b/lib/utils/interpret-conditionals.test.js
@@ -82,3 +82,28 @@ test('interpretConditionals removes empty lines from removed conditionals', t =>
   const log = fastCopy(logData)
   t.assert.strictEqual(interpretConditionals('{msg}\n{if foo}{foo}{end}\n{level}', log), '{msg}\n{level}')
 })
+
+test('interpretConditionals does not interpret `$` sequences in the property value', t => {
+  const log = { msg: 'a $& b' }
+  t.assert.strictEqual(interpretConditionals('{if msg}{msg}{end}', log), 'a $& b')
+})
+
+test('interpretConditionals preserves every `$` replacement pattern in the property value', t => {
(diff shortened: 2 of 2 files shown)


kilisamemarisaaa (approved):
Verified independently at `f033c3802940087fbc4b8823ef419a2f0694e77f`.

Using a string search removes the regular-expression interpretation of the key, and the function replacer preserves dollar-ampersand, double-dollar, dollar-backtick, and dollar-apostrophe replacement sequences in logged values. I also checked repeated placeholders, numeric and object values, escaped-dot keys, and keys containing `$`; those compatibility cases retain the existing coercion and lookup behavior.

Validation on Windows with Node 24.12.0:

- affected test file: 18/18 passed
- `npm run lint`: passed
- `npm run test:types`: 9/9 tests and 66/66 assertions passed; package type check reported no problems
- `borp --check-coverage` reported every test file as passed, but the runner did not terminate after its CLI child-process tests on this Windows host, so I interrupted it rather than claiming a completed full run

I found no blocking issue in the change.

